PR #2, opened 2015-12-04
Event response buttons and dropdown

robbiecarlton:


src/components/Post.js
```diff
@@ -151,18 +151,82 @@ export default class Post extends React.Component {
...
   let {
     post, image, comments, commentsExpanded,
-    commentingDisabled, onCommentCreate, communities
+    commentingDisabled, onCommentCreate, communities, currentUser
   } = props
   let description = present(sanitize(post.description))
   let attachments = filter(post.media, m => m.type !== 'image')
   if (!comments) comments = []
 
+  var yeses = filter(post.responders, er => er.response === 'yes')
```
robbiecarlton: @levity - there's so much code specifically for the event responses, I'm tempted to pull it out into it's own component. What are your thoughts?

levity: Making components smaller seems to be a good rule of thumb in React.

robbiecarlton:
screen shot 2015-12-04 at 3 58 05 pm
@levity ok to merge?

src/components/RSVPControl.js
```diff
@@ -0,0 +1,86 @@
...
+          <button type='button' onClick={eventResponseClicked('yes')} className={'rsvp-yes btn btn-default ' + (eventResponse === 'yes' ? 'active' : '')}>
+            Going
+            {yeses.length > 0
+            ? ' (' + yeses.length + ')'
+            : null}
+          </button>
+          <button type='button' onClick={eventResponseClicked('maybe')} className={'rsvp-maybe btn btn-default ' + (eventResponse === 'maybe' ? 'active' : '')}>
+            Maybe
+            {maybes.length > 0
+            ? ' (' + maybes.length + ')'
```
levity: can you use string interpolation (backticks) for this sort of thing instead of +?

src/components/Post.js
```diff
@@ -93,7 +94,7 @@ export default class Post extends React.Component {
 
     return <div className={classes} onClick={this.expand}>
       <div className='header'>
-        <Avatar person={person}/>
+        <Avatar person={person} />
```
levity: ?

src/components/RSVPControl.js
```diff
@@ -0,0 +1,86 @@
...
+export default class RSVPControl extends React.Component {
+  static propTypes = {
+    currentUser: object,
+    post: object,
+    dispatch: func
+  }
+
+  render () {
+    var props = this.props
+    let { post, currentUser } = props
```
levity: why not combine these two into `let { ... } = this.props`?

src/components/RSVPControl.js
```diff
@@ -0,0 +1,86 @@
...
+
+    var yeses = filter(post.responders, er => er.response === 'yes')
```
levity: i'm seeing a lot of stuff here repeated 3 times -- can you just pull it out into a function and call it 3 times?

src/components/RSVPControl.js
levity: btw, if you are just displaying nothing when the condition is false, you can do `foo && 'hello'` instead of `foo ? 'hello' : null`

src/components/RSVPControl.js
```diff
@@ -0,0 +1,86 @@
...
+            : null}
+          </button>
+        </div>
+
+        {post.responders.length > 0 && <div className='responses'>
+
+          <Dropdown className='responses-dropdown' toggleChildren={<span>See Responses</span>}>
+            {responderList.map(personOrHeader => <li key={personOrHeader.id} className={personOrHeader.header ? 'header' : ''}>
+              {personOrHeader.header
+              ? <span className='header'>{personOrHeader.header}</span>
```
levity: when you split a ternary expression across multiple lines, can you indent the ? and : please?

src/reducers/posts.js
```diff
@@ -13,6 +13,18 @@ const normalizeUpdate = (post, params) => {
...
+const changeEventResponse = (post, response, user) => {
+  if (!user) return
+
+  var meInResponders = findWhere(post.responders, {id: user.id})
+  var responders = without(post.responders, meInResponders)
+
+  if (!meInResponders || meInResponders.response !== response) {
+    responders.push({id: user.id, name: user.name, avatar_url: user.avatar_url, response: response})
+  }
+  return Object.assign({}, post, { responders: responders })
```
levity: you can use object spread syntax here: `{...post, ...{responders}}`

src/reducers/posts.js
```diff
@@ -31,6 +43,13 @@ export default function (state = {}, action) {
       let { params, id } = meta
       let post = state[id]
       return {...state, [id]: normalizeUpdate(post, params)}
+    case CHANGE_EVENT_RESPONSE_PENDING:
+      id = meta.id
+      post = state[id]
+      var { response, user } = meta
+      return {...state, [id]: changeEventResponse(post, response, user)}
+    case CHANGE_EVENT_RESPONSE:
+      return state
```
levity: no need to explicitly return state here

src/components/RSVPControl.js
```diff
@@ -0,0 +1,86 @@
...
+              : <span className='person'>
+                <Avatar person={personOrHeader} /> <Link className='person' to={`/u/${personOrHeader.id}`}><span>{personOrHeader.name}</span></Link>
```
levity: it would be cleaner if, instead of a one-dimensional list of mixed-up headers and people, you had a list of lists:

```
[
  ['yes', [... people ...]],
  ['no', [... people ...]],
  ['maybe', [... people ...]]
]
```

src/components/RSVPControl.js
```diff
@@ -0,0 +1,86 @@
...
+    var eventResponseClicked = function (response) {
+      return e => {
+        props.dispatch(changeEventResponse(post.id, response, currentUser))
+      }
+    }
+
+    return (
+      <div className='rsvp-controls post-section'>
+        <div className='btn-group buttons'>
+          <button type='button' onClick={eventResponseClicked('yes')} className={'rsvp-yes btn btn-default ' + (eventResponse === 'yes' ? 'active' : '')}>
```
levity: for `className`, look at the way i've been using the `classnames` module

src/components/RSVPControl.js
robbiecarlton: I also need the props var for later in eventResponseClicked (can't use this.props because this is not bound to the same thing anymore)
Here I could do this instead: 

```
var props = this.props
let { post, currentUser } = this.props
```

your call

levity: you only need props for dispatch, so i would do:

``` js
let { post, currentUser, dispatch } = this.props
// ...
var eventResponseClicked = response => e => dispatch(...)
```

btw, you can also skip the function nesting and just do something like this:

``` js
var pickResponse = choice => dispatch(...)

<button onClick={() => pickResponse('yes')}>yes</button>
```

robbiecarlton: thanks, that's much nicer

src/components/RSVPControl.js
robbiecarlton: thanks, that's better. I was in the trance of the version I did for angular, knew there was a better way to do it here.

robbiecarlton:
@levity how's this?
I ended up doing a list of objects over list of lists for the event responses, it seemed more readable.

levity:
better! you can also pull those little functions out of the main render method. (renderRSVPButton will just need additional parameters instead of depending on closing over local scope.)

levity:
also, corollary to the general principle of making smaller components, we also want to make components depend on smaller amounts of state. so in this case i would pass only `responders` and `onPickResponse` to RSVPControl, so that the dependence on `post.id`, `currentUser`, and `dispatch` can stay in the owning component. and define `onPickResponse` in the owning component as `choice => dispatch(changeEventResponse(post.id, choice, currentUser)`

robbiecarlton:
I still need currentUser for rendering the buttons (knowing which should be active) I could pull that out into the owning component as well, so instead of passing currentUser, i just pass currentUserResponse. Do you prefer that?

levity:
yeah, it seems conceptually sound for the control to know about the current response

robbiecarlton:
@levity can I merge?

levity:
yep

PR #4, opened 2015-12-09
User Settings

robbiecarlton:


src/containers/user/UserSettings.js
```diff
@@ -0,0 +1,227 @@
...
+    edited[field] = currentUser[field]
+    this.setState({editing: {...editing, [field]: true}})
+  }
+
+  cancelEdit (field) {
+    let { editing } = this.state
+    this.setState({editing: {...editing, [field]: false}})
+  }
+
+  toggle (field) {
```
robbiecarlton: @levity - do you have a preference for doing this using 3 methods like so with a lot of repetition, versus one long harder to read method like this:
https://github.com/Hylozoic/hylo-redux/commit/15ae157ab614c0147ba852265f6ff377511d8533

robbiecarlton: also, do you still think updateUserSettings and updateUserSettingsEditor should be one action, given that they have different signatures now?

levity: the best refactoring i see would be something like:

``` js
update (field, value) {
  let updatedUser = {...currentUser, [field]: value}
  dispatch(...)
}

updateSetting (field, value) (
  let updatedSettings = makeUpdatedSettings(field, value)
  update('settings', updatedSettings)
}
```

...and then just handle the toggle cases by writing out `update('foo', !currentUser.foo)` in the template. this can be pretty concise with liberal use of `let { ... } = currentUser` and `let { ... } = settings`. or write two very small toggle methods that wrap the methods above.

btw, note redundant syntax: 
`{...currentUser, ...{[field]: value}}` =>
`{...currentUser, [field]: value}`

levity: i think `updateUserSettingsEditor` is redundant because you can do the same things in the reducers by listening to `UPDATE_USER_SETTINGS_PENDING` instead.

robbiecarlton: ah. this is nice. Thanks!

robbiecarlton:
@levity good to merge?

Couple of things, I took out the 'join a community' link when the user doesn't have any communities. I'll add it back in when we have join community flow in redux.

Could you check out the latency of the check boxes in settings. Right now they're waiting on an action roundtrip, and it's just noticeable. I'm wondering if we should also update them locally...

src/reducers/index.js
```diff
@@ -97,6 +108,21 @@ export default combineReducers({
...
     }
 
+    if (type === UPDATE_USER_SETTINGS_PENDING) {
+      return {
+        ...state,
+        current: {...state.current, ...meta.params}
+      }
+    }
+
+    if (type === LEAVE_COMMUNITY_PENDING) {
```
levity: you need a corresponding `type === LEAVE_COMMUNITY && error` case to undo this, yeah?

src/containers/user/UserSettings.js
```diff
@@ -0,0 +1,270 @@
...
+        break
+    }
+  }
+
+  render () {
+    let { currentUser } = this.props
+    let { editing, edited, errors, expand1, expand2, expand3 } = this.state
+
+    return <div id='user'>
+      <div className='settings'>
```
levity: oh man -- i really would rather not carry over all this old HTML... oh well, i guess we can clean this up later

src/containers/user/UserSettings.js
```diff
@@ -0,0 +1,270 @@
...
+        <div className='section-label' onClick={() => this.setState({expand1: !expand1})}>
+          Account
+          <i className={cx({'icon-down': expand1, 'icon-right': !expand1})}></i>
+        </div>
+        {expand1 && <div className='section email'>
```
levity: let's at least use meaningful names instead of `expandN`

src/containers/user/UserSettings.js
```diff
@@ -0,0 +1,270 @@
+import validator from 'validator'
+import React from 'react'
+import { connect } from 'react-redux'
+import { prefetch } from 'react-fetcher'
+import cx from 'classnames'
+const { func, object } = React.PropTypes
+import { fetchCurrentUser, updateUserSettings, leaveCommunity } from '../../actions'
+import A from '../../components/A'
+
+const formatDate = (date) => {
```
levity: move this to util/text

levity:
lemme know if anything hairy shows up when trying to resolve conflicts

src/reducers/index.js
```diff
@@ -97,6 +115,21 @@ export default combineReducers({
       }
     }
 
+    if (type === UPDATE_USER_SETTINGS_PENDING) {
```
robbiecarlton: Any objection to me putting the PENDING cases and the error cases each in their own switch statements, instead of the ifs?

robbiecarlton:
@levity can I merge?

css/index.less
```diff
@@ -893,6 +893,57 @@ ul.tags {
...
+
+  .section-item {
+    .make-row;
+    padding: 15px 0px;
+
+    input[type=checkbox] {
+      .round-checkbox();
+    }
+
+  .summary {
```
levity: indentation's off here

css/index.less
```diff
@@ -893,6 +893,57 @@ ul.tags {
...
+    .meta-text
+  }
+
+  }
+
+  label {
+    font-family: "proxima-nova", sans-serif;
```
levity: we shouldn't need to use font-family anywhere except in mixins in type.less

css/mixins.less
```diff
@@ -99,6 +99,35 @@
...
 
+.round-input {
+  -webkit-appearance: none;
+  border: 1px solid @button-border;
+  border-radius: 25px;
+  outline: none;
+  margin: 0;
+  display: inline-block;
+  position: relative;
+  .proxima-nova(100);
```
levity, line 110: i think this is the same as .proxima-nova-light?

src/reducers/index.js
```diff
@@ -82,19 +86,47 @@ export default combineReducers({
   },
 
   people: (state = {}, action) => {
```
levity: this is ready to be broken out into its own file and tested

src/actions/index.js
```diff
@@ -199,3 +203,19 @@ export function changeEventResponse (id, response, user) {
...
+    payload: {api: true, params, path: `/noo/user/${params.id}`, method: 'POST'},
+    meta: {params, prevProps}
+  }
+}
+
+export function leaveCommunity (communityId, prevProps) {
+  return {
+    type: LEAVE_COMMUNITY,
+    payload: {api: true, path: `/noo/membership/${communityId}`, method: 'DELETE'},
+    meta: {communityId: communityId, prevProps}
```
levity: you can just do `{communityId, ...}` instead of `{communityId: communityId, ...}`

css/globals.less
```diff
@@ -10,6 +10,7 @@
 @hylo-teal: #4fb19f;
 @hylo-orange: #e79336;
 @hylo-green: #52b065;
+@button-border: #B3B3B3;
```
levity: where's this from? bootstrap, or our own styles from hylo-frontend? if the latter, let's just use `@light-gray-border` instead, which matches bootstrap

robbiecarlton:
@levity can I merge?

levity:
:+1:!

PR #6, opened 2015-12-10
Community settings

robbiecarlton:


robbiecarlton:
@levity How's it look?
The one thing I wanted to point out that's different from the angular version is: in this version I prefetch all the users in the community, for the add moderator and change leader inputs.
This is to keep the code simpler, but it's a little more server and browser load. Let me know if you want me to do it using the autocomplete api

levity:
Yeah, let's use the autocomplete API. That's a lot of unnecessary data to
schlep if someone just wants to change the banner image or something. And
it's weird to see connectedListProps and so forth being used when the
component doesn't have a paginated list. Should be straightforward to use
what I built for @-mentions instead.

Having all those setState calls where we're changing nested data seems like
an antipattern. Not exactly sure what to do but it probably involves
splitting out logic into more components. Possibly a wrapper component for
each item (or group of items being edited together).

That's what I got from scanning the PR on my phone. Will give it a more
careful look when I'm at a computer tomorrow, the one day I'm back in SF
between Taiwan and Halifax.

robbiecarlton:
Switched to using autocomplete
If we have wrapper components for each group of setting items, does that mean we would be using connect with each wrapper component? Otherwise we end up with all that nested state logic in the main container anyway.

PR #7, opened 2015-12-22
Projects join flow and contributors

robbiecarlton:


robbiecarlton:
@levity - could you take a look at this as well if you get the chance, just to see if I'm on the right track

src/components/PersonListItem.js
```diff
@@ -0,0 +1,73 @@
...
+
+@connect(({ people }, { personId }) => {
+  console.log('Connect', people)
+  console.log('PersonId', personId)
+  return { person: people[personId] }
+})
+export default class PersonListItem extends React.Component {
+  static propTypes = {
+    person: object,
+    personId: string
```
levity: it's ok to just use `id` here instead of `personId` since its meaning is apparent from context

src/reducers/projects.js
```diff
@@ -13,6 +22,14 @@ export default function (state = {}, action) {
       return {...state, [payload.id]: payload}
     case UPDATE_PROJECT:
       return {...state, [id]: {...state[id], ...params}}
+    case JOIN_PROJECT_PENDING:
+      var contributors
+      if (currentUser) {
+        contributors = union(state[id].contributors, [currentUser])
```
levity: you can use spread notation here: `[...state[id].contributors, currentUser]`

levity:
so far, so good :+1:

src/components/PersonListItem.js
robbiecarlton: I'm actually in the process of rewriting this so it just takes a person and doesn't need to be connected, and the contributors tab fetches and

robbiecarlton:
@levity good to merge?

src/components/PersonChooser.js
```diff
@@ -24,7 +24,7 @@ export default class PersonChooser extends React.Component {
   handleInput = event => {
     var value = event.target.value
     let { dispatch, typeaheadId, communityId } = this.props
-    dispatch(typeahead(value, typeaheadId, communityId))
+    dispatch(typeahead(value, typeaheadId, { communityId, type: 'people' }))
```
levity: no spaces between curly brackets here please

src/reducers/projects.js
levity: bump

levity:
looks good

PR #8, opened 2016-01-07
Make choices a prop of TagInput

robbiecarlton:


robbiecarlton:
@levity Like this?

src/components/TagInput.js
```diff
@@ -8,21 +8,15 @@ export default class TagInput extends React.Component {
...
-    super(props)
-    this.state = {choices: []}
-  }
-
   handleInput = event => {
     var value = event.target.value
-    var choices = this.props.getChoices(value) || []
-
-    this.setState({choices: choices})
+    this.props.handleInput(value)
```
levity: At this point you might as well remove the local handleInput method entirely

src/components/PostEditor.js
```diff
@@ -166,20 +171,24 @@ export default class PostEditor extends React.Component {
...
       !contains(communities, c.id)
 
-    return filter(currentUser.memberships.map(m => m.community), match)
+    this.setState({communityChoices: filter(currentUser.memberships.map(m => m.community), match)})
   }
 
   render () {
     let { expanded, post, postEdit, dispatch, project } = this.props
     let { name, description, communities, type, location } = postEdit
+    let { communityChoices } = this.state
```
levity: The only thing that needs to be in state is the term itself -- you can calculate the communities that correspond to the term in the render method rather than storing them in state

Optionally of course you can then move the term to the app state

levity:
Looks good oyherwise

robbiecarlton:
@levity how's this?

levity:
:+1:

PR #9, opened 2016-01-14
Notifications

robbiecarlton:


robbiecarlton:
screen shot 2016-01-14 at 6 11 06 pm
@levity how's this look?

levity:
so far so good -- bump the "Mark all as read" button up to be in line with
the header, like we do on the All Posts page

src/containers/Notifications.js
```diff
@@ -0,0 +1,151 @@
...
+import cx from 'classnames'
+import ScrollListener from '../components/ScrollListener'
+import A from '../components/A'
+import truncate from 'html-truncate'
+import { present, humanDate } from '../util/text'
+const { array, bool, func, number, object } = React.PropTypes
+
+@prefetch(({ dispatch }) => dispatch(fetchActivity(20, 0)))
+@connect(state => ({
+  activities: sortBy(values(state.activities), ['created_at']).reverse(),
```
levity: shouldn't these be arriving from the API sorted already?

levity: oh never mind, you're working around the case where a new activity shows up during pagination?

levity: wait, that's not going to be sufficient, because you're not going to get the new ones. crap, this is a bug with all of our pagination. the way i've solved this in the past (don't know how i overlooked this here) is to pass the id or timestamp of the item at the front, and then add a clause on the backend to include only items with id/timestamp equal to or less than that one

src/containers/Notifications.js
```diff
@@ -0,0 +1,151 @@
...
+    let { dispatch } = this.props
+    dispatch(markAllActivitiesRead())
+  }
+
+  visit = activity => {
+    let { dispatch } = this.props
+    let path = `/p/${activity.post.id}`
+    if (activity.unread) {
+      dispatch(markActivityRead(activity.id))
+      .then(action => {
```
levity: in this situation you might as well not even wait for `.then`, because the second action doesn't depend on the first. might be slightly more responsive?

src/reducers/index.js
```diff
@@ -266,6 +270,51 @@ export default combineReducers({
...
+          }, {})
+        }
+      case MARK_ACTIVITY_READ:
+        return {
+          ...state,
+          [meta.activityId]: {...state[meta.activityId], unread: false}
+        }
+      case MARK_ALL_ACTIVITIES_READ_PENDING:
+        return mapValues(state, activity => ({...activity, unread: false}))
+      case THANK_PENDING:
```
levity: the commentsByPost reducer should handle this too

src/containers/Notifications.js
robbiecarlton: actually, this was because I was storing the activities as an object with their ids as keys. I rewrote to store them in an array. Not sure which I like better, but see https://github.com/Hylozoic/hylo-redux/commit/dbc5fe992094133b3d678bf1f57e3e898d8a0e7b for the array version

seems like the pagination fix is a bigger update. Let me know if you want me to have a crack at it.

robbiecarlton:
@levity: This now incorporates the 'comments store' pr: https://github.com/Hylozoic/hylo-redux/pull/10

Let me know if you think it's ready to merge...

test/reducers/comments.test.js
```diff
@@ -49,4 +51,71 @@ describe('comments', () => {
...
+        '1': {id: '1', comment_text: 'foo', thanks: [{comment_id: '1', thanked_by_id: '10'}]},
+        '2': {id: '2', comment_text: 'bar', thanks: []}
+      }
+
+      expect(comments(state, action)).to.deep.equal(expectedState)
+    })
+  })
+
+  describe('on THANK_PENDING', () => {
+    it('removes a thanks', () => {
```
levity: this test should make sure it only passes if action.error is true

levity:
:+1:

PR #11, opened 2016-01-20
"join community" link and page

robbiecarlton:


src/containers/community/CommunityJoin.js
```diff
@@ -0,0 +1,63 @@
...
+  constructor (props) {
+    super(props)
+    this.state = {codeEmpty: true}
+  }
+
+  codeChanged = event => {
+    let { dispatch } = this.props
+    let { value } = event.target
+    if (isEmpty(value.trim())) {
+      this.setState({codeEmpty: true})
```
robbiecarlton: How do you feel about this use of state?
the alternative would be to do a round trip through the communityValidation store, which would simplify the code here, at the expense of more code in the reducer. I could go either way...

src/containers/community/CommunityJoin.js
```diff
@@ -0,0 +1,63 @@
...
+      dispatch(resetCommunityValidation('beta_access_code'))
+    } else {
+      this.setState({codeEmpty: false})
+      dispatch(validateCommunityAttribute('beta_access_code', value, 'unique'))
```
levity: you want 'exists' here instead of 'unique'

src/containers/community/CommunityJoin.js
levity: i think state is fine. even better to skip `setState` entirely and just check `this.refs.code.value` in `render()`.

robbiecarlton: refs are not available in the initial render, so I think it's simpler to use state here

src/containers/community/CommunityJoin.js
robbiecarlton: hah. That just solved the whole codeEmpty state issue

robbiecarlton:
good to merge?

src/containers/community/CommunityJoin.js
```diff
@@ -0,0 +1,54 @@
...
+    let { dispatch, codeInvalid } = this.props
+
+    if (codeInvalid) {
+      return
+    }
+    let code = this.refs.code.value
+    dispatch(joinCommunityWithCode(code))
+    .then(action => {
+      if (action.error) {
+        console.log('TODO - display error message')
```
levity: all the errors are being stored in the `errors` subtree of state, so you can use that to get an error message into props via `connect` -- follow the error presentation style of Login and Signup

robbiecarlton:
like this? Had to add a RESET_ERROR action to clear the error message

src/containers/community/CommunityJoin.js
```diff
@@ -0,0 +1,61 @@
...
+import { get, isEmpty } from 'lodash'
+import { JOIN_COMMUNITY_WITH_CODE, resetError, joinCommunityWithCode, navigate, resetCommunityValidation, validateCommunityAttribute } from '../../actions'
+import cx from 'classnames'
+
+@connect(({communityValidation, errors}) => {
+  var error
+  if (errors[JOIN_COMMUNITY_WITH_CODE]) {
+    error = 'We couldn\'t find that code, please check it and try again'
+  }
+  let codeInvalid = !get(communityValidation, 'beta_access_code.exists')
```
levity, line 20: i think you need to explicitly check `=== false`

src/containers/community/CommunityJoin.js
```diff
@@ -0,0 +1,61 @@
...
+      dispatch(resetCommunityValidation('beta_access_code'))
+    } else {
+      dispatch(validateCommunityAttribute('beta_access_code', value, 'exists'))
+    }
+  }
+
+  submit = () => {
+    let { dispatch, codeInvalid } = this.props
+
+    if (codeInvalid) {
```
levity: do this on a single line pls

src/containers/community/CommunityJoin.js
```diff
@@ -0,0 +1,61 @@
...
+      return
+    }
+    let code = this.refs.code.value
+    dispatch(joinCommunityWithCode(code))
+    .then(action => {
+      if (action.error) {
```
levity: here too

levity: actually this can be shortened to `.then(({ error }) => error || dispatch(...))`

src/containers/community/CommunityJoin.js
robbiecarlton, line 20: Actually don't want to explicitly check false. Both false and undefined should set codeInvalid to true. This is what let's me omit the codeEmpty stuff entirely.

src/containers/community/CommunityJoin.js
```diff
@@ -0,0 +1,61 @@
...
+  static propTypes = {
+    dispatch: func,
+    codeInvalid: bool,
+    error: string
+  }
+
+  codeChanged = event => {
+    let { dispatch } = this.props
+    let { value } = event.target
+    dispatch(resetError(JOIN_COMMUNITY_WITH_CODE))
```
levity: only dispatch if there is an error

src/containers/community/CommunityJoin.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+import { connect } from 'react-redux'
+const { func, bool, string } = React.PropTypes
+import { get, isEmpty } from 'lodash'
+import { JOIN_COMMUNITY_WITH_CODE, resetError, joinCommunityWithCode, navigate, resetCommunityValidation, validateCommunityAttribute } from '../../actions'
```
levity: break this into multiple lines, viz:

```
import {
  a,
  b
} from 'foo'
```

src/containers/community/CommunityJoin.js
levity, line 20: aha, gotcha

robbiecarlton:
@levity good to merge now?

levity:
:+1:

PR #16, opened 2016-02-10
Network admin

robbiecarlton:


robbiecarlton:
@levity this is currently missing validation of slug, but I'd like to merge now so that Ed can use it tonight, while I work on validation

robbiecarlton:
Actually, don't worry about it, the validation version will be ready

PR #18, opened 2016-02-12
(wip) Networks

robbiecarlton:


src/containers/network/AboutNetwork.js
```diff
@@ -1,3 +1,17 @@
 import React from 'react'
-const AboutNetwork = props => <span />
-export default AboutNetwork
+import { connect } from 'react-redux'
+import { markdown } from '../../util/text'
+const { object } = React.PropTypes
+
+@connect((state, { params }) => ({network: state.networks[params.id]}))
+export default class AboutNetworks extends React.Component {
```
levity, line 7: see if you can stick with stateless components unless absolutely necessary. they're more performant (according to the React docs) and help us avoid antipatterns. 

decorators just use function composition, so you can do:

``` js
const MyComponent = connect(...)(props => { ... })
```

or

``` js
import { compose } from 'redux'
const MyComponent = compose(
  prefetch(...),
  connect(...)
)(props => { ... })
```

you can declare propTypes thus:

``` js
MyComponent.propTypes = { ... }
```

src/containers/network/NetworkMembers.js
```diff
@@ -1,3 +1,67 @@
...
+
+@prefetch(({ dispatch, params: { id }, query }) => dispatch(fetch(subject, id, query)))
+@connect((state, { params: { id }, location: { query } }) => {
+  return {
+    ...connectedListProps(state, {subject, id, query}, 'people'),
+    network: state.networks[id],
+    currentUser: state.people.current
+  }
+})
+export default class NetworkMembers extends React.Component {
```
levity, line 22: this can be stateless too; you can define `loadMore` and `updateQuery` inside `render`

src/containers/network/AboutNetwork.js
levity, line 7: oh i wrote all that before i noticed you'd already done it for `NetworkPosts` :)

PR #19, opened 2016-02-15
Image upload bug fix

robbiecarlton:
This fixes an issue in community and network creation where, if you upload an image and then immediately save the community/network, before the UPLOAD_IMAGE action has fired, you lose the image. 
Now you cannot save while UPLOAD_IMAGE is pending

### DIFF
--- a/src/containers/community/CommunityEditor.js
+++ b/src/containers/community/CommunityEditor.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux'
 import { some, get, omit, toPairs } from 'lodash'
 import {
   CREATE_COMMUNITY,
+  UPLOAD_IMAGE,
   createCommunity,
   navigate,
   resetCommunityValidation,
@@ -36,6 +37,7 @@ const categories = {
   let validating = some(communityValidation.pending)
   let { community, errors } = communityEditor
   let saving = pending[CREATE_COMMUNITY]
+  let uploadingImage = !!pending[UPLOAD_IMAGE]
 
   if (!errors) errors = {}
   errors.nameUsed = get(communityValidation, 'name.unique') === false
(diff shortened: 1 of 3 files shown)


src/containers/community/CommunityEditor.js
```diff
@@ -36,6 +37,7 @@ const categories = {
   let validating = some(communityValidation.pending)
   let { community, errors } = communityEditor
   let saving = pending[CREATE_COMMUNITY]
+  let uploadingImage = !!pending[UPLOAD_IMAGE]
```
robbiecarlton, line 40: I did this in the most verbose way, by adding a new prop. Another way would be to roll it in to `saving`, maybe rename `saving`
ie
`let pending = pending[CREATE_COMMUNITY] || pending[UPLOAD_IMAGE]`
figured it's better to have it more articulated.

src/containers/community/CommunityEditor.js
```diff
@@ -313,7 +316,7 @@ export default class CommunityEditor extends React.Component {
           {errors.server || 'The information you provided has some errors; please see above.'}
         </p>}
         <button type='button' onClick={this.submit} disabled={disableSubmit}>
-          {validating || saving ? 'Please wait...' : 'Save'}
+          {validating || saving || uploadingImage ? 'Please wait...' : 'Save'}
```
robbiecarlton, line 319: is there any reason this line is not just
`{disableSubmit ? 'Please wait...' : 'Save'}`

levity:
:+1:

src/containers/community/CommunityEditor.js
levity, line 40: oh actually that's fine too. we could just call it 'pending' to indicate that it's not just because saving is taking place

src/containers/community/CommunityEditor.js
levity, line 319: note that `disableSubmit` is true if there are any errors. in this case the text shouldn't read "Please wait" because there's nothing to wait for

robbiecarlton, line 319: got it

PR #23, opened 2016-02-22
original Network Admin PR, aka Revert "Revert "Network Admin""

robbiecarlton:
Reverts Hylozoic/hylo-redux#22

### DIFF
--- a/src/reducers/networks.js
+++ b/src/reducers/networks.js
@@ -1,7 +1,10 @@
 import { debug } from '../util/logging'
+import { map } from 'lodash'
 import {
   CREATE_NETWORK,
-  FETCH_NETWORK
+  FETCH_NETWORK,
+  FETCH_COMMUNITIES,
+  UPDATE_NETWORK_PENDING
 } from '../actions'
 
 export default function (state = {}, action) {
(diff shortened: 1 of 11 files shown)


robbiecarlton:
@levity ok, this is ready for you to look at.

src/reducers/networks.js
```diff
@@ -10,17 +13,25 @@ export default function (state = {}, action) {
...
     case CREATE_NETWORK:
       return {
         ...state,
         [payload.slug]: payload
       }
+    case FETCH_COMMUNITIES:
+      let newCommunities = (state[id].communities || []).concat(map(payload.communities, 'id'))
+      return {...state, [id]: {...state[id], communities: newCommunities}}
+    case UPDATE_NETWORK_PENDING:
+      console.log('networks reducer, UNP', {id, params: meta.params})
```
levity: stray console.log here

levity:
looks good otherwise. it does feel like it's lacking some tests, but i think i need to figure out a way to simulate a client-side environment in the test suite so we can write end-to-end tests where we click buttons and expect state to change accordingly

robbiecarlton:
Agreed, lacking tests, and end to end tests especially. Might be fun to pair on that some...

PR #25, opened 2016-02-24
add Delete button to community settings

robbiecarlton:


robbiecarlton:
@levity What do you think?

src/reducers/people.js
```diff
@@ -62,6 +63,14 @@ export default function (state = {}, action) {
         ...state,
         current: {...state.current, memberships}
       }
+    case UPDATE_COMMUNITY_SETTINGS_PENDING:
```
robbiecarlton, line 66: I couldn't find a neater way of doing this than catching the action in both the communities and people reducers. But I'm a little suspicious of the duplication...

levity, line 66: nah, it's fine. we should be normalizing the data under currentUser so that memberships just has a list of ids, or break memberships itself into its own section of the store, but even then you would have to remove the membership.

levity:
:+1:

PR #26, opened 2016-02-25
Update new_notification_count from /noo/live-status

robbiecarlton:


src/server/appHandler.js
```diff
@@ -82,8 +82,10 @@ function renderApp (res, renderProps, history, store) {
     )
 
     let state = store.getState()
+    let unreadCount = get(state, 'people.current.new_notification_count')
 
     return React.createElement(Html, {
+      pageTitle: 'Hylo' + (unreadCount && ` (${unreadCount})`),
```
robbiecarlton: this is not updating when the store updates, because it's outside Redux. Any thoughts on how to get this updating? One possibility would be doing it with JS in eg the App component, but I'm hoping you have a better way...

levity: http://stackoverflow.com/questions/30280816/whose-job-is-it-in-flux-react-to-set-the-page-title

This approach feels ok to me -- set the title in the store, add the value
to the connect decorator in the App component, then create a
componentWillReceiveProps method that just changes the page title with
vanilla JavaScript when the prop changes.

robbiecarlton:
@levity I ended up creating a new component, to keep App stateless. Don't know if that's important. Lmk if you want me to move it into App

levity:
that's fine too

PR #28, opened 2016-02-26
(WIP) Show live notification for new posts in current feed

robbiecarlton:
@levity How's this for the general approach? Specifically, doing all the logic in connectedListProps

### DIFF
--- a/src/components/PostList.js
+++ b/src/components/PostList.js
@@ -11,6 +11,7 @@ class PostList extends React.Component {
   static propTypes = {
     posts: array,
     loadMore: func,
+    newPosts: func,
     pending: bool,
     editingPostIds: array
   }
@@ -47,7 +48,7 @@ class PostList extends React.Component {
   }
 
   render () {
-    let { posts, editingPostIds, pending, loadMore } = this.props
+    let { posts, editingPostIds, pending, loadMore, newPosts } = this.props
     let { project } = this.context
 
     if (!pending && posts.length === 0) {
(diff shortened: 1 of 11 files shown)


robbiecarlton:
@levity - functionality is done. 
Not sure about the best UX for the new posts notification. You have any thoughts?

src/components/PostList.js
```diff
@@ -56,6 +57,9 @@ class PostList extends React.Component {
 
     return <ul className='posts'>
       {pending && <li className='loading'>Loading...</li>}
+      {newPosts && <li>
+        <div className='post new-posts'>New Posts are available. <a onClick={newPosts}>Refresh</a></div>
```
levity, line 61: you'll definitely want to animate this (see my use of `velocity-react` elsewhere in the code)

src/reducers/stagedPostsByQuery.js
```diff
@@ -0,0 +1,22 @@
...
+} from '../actions'
+
+export default function (state = {}, action) {
+  if (action.error) return state
+  if (!get(action, 'meta.staged')) return state
+
+  let { type, payload, meta } = action
+  switch (type) {
+    case FETCH_POSTS:
+      return appendUniq(state, meta.cacheId, map(payload.posts, 'id'))
```
levity, line 16: I think this should be just replacing `state[meta.cacheId]` instead of appending, no? e.g. so that we don't get duplicates if the staged fetch runs multiple times

levity:
What do you think about this approach now that you've written it? 

I can't help but feel like it's not quite right, that it's scattered across a bunch of different locations, and muddies the logic of `fetchPosts`. I think, now that I see the code (and I likely would have come to the same conclusion had I written it), that it would be clearer and more future-proof if instead:
- there's a new API endpoint called, say, `PostController.checkFreshness`
- which you send as parameters `subject`, `id`, `query`, and maybe the `id` and `updated_at` of each post that you already have
- and returns true or false
- this boolean triggers the showing of the "New Posts" button" (which should be hidden if any of the query parameters change)
- and if the button is pressed, just clear the cache for the current `cacheId` and fetch as if it were the first time

I think that would be easier to follow on the front-end, and also leaves a lot of flexibility on the back-end to change the criteria by which we decide to trigger the New Posts notification, where the current approach hardcodes the rule "change if post ID's are different". For example it could trigger if one of the existing posts has a new comment.

robbiecarlton:
agreed with everything. Working on that version now

PR #38, opened 2016-04-09
Hashtag autocomplete

levity:
overhaul the way tags (hashtags & mentions) are added in rich text editing


levity:
@robbiecarlton this is what i've been working on. At this point it's more or less ready to merge into new-look, as the @-mention behavior is the same as before, except that Enter doesn't work to select an autocomplete option so you have to use Tab instead, and a few other little things i'm tracking at https://app.asana.com/0/8055896560428/109452871221320. Just wanted to get your eyes on it for a bit for sanity checking. Feel free to ask questions.

PR #39, opened 2016-04-15
Tagging

robbiecarlton:


src/util/linkify.js
```diff
@@ -3,43 +3,50 @@ import linkifyString from 'linkifyjs/string'
...
+  ? `/c/${slug}/tag/${tagName}`
+  : `/tag/${tagName}`
+}
+
 // this handles old-style hashtags, which aren't wrapped in tags
 var linkifyjsOptions = {
   formatHref: function (value, type) {
     if (type === 'hashtag') {
-      return '/search?q=%23' + value.substring(1)
+      return '/tag/' + value.substring(1)
```
robbiecarlton: I got stuck here on how to get the community slug into linkifyjs to create the full tag.
I'm at the point where it seems like I may have to roll my own function that turns the hashtags into links, but wanted to run it past you first.

levity: you have to define the options inside `linkify` now instead of having them be constant.

robbiecarlton: ah! thanks

robbiecarlton:
@levity - good to merge?

src/util/text.js
```diff
@@ -24,20 +24,24 @@ export function sanitize (text) {
...
   // make links and hashtags
-  text = linkify(text)
+  text = linkify(text, opts.slug)
 
   if (opts && opts.maxlength) text = truncate(text, opts.maxlength)
 
   return text
 }
 
+export function formatPostTitle (text, slug) {
```
levity: this is basically `present` without wrapping in a `<p>` tag -- can you do that instead, i.e. add a `skipWrap` option to `present`, or add a `wrap` option? it feels wrong to have a utility file know about posts

levity: hmm... it feels wrong to have a util file mention posts

src/components/Post.js
```diff
@@ -189,11 +194,18 @@ export class CommentSection extends React.Component {
...
+    dispatch: func,
+    community: object,
+    currentUser: object
+  }
 
   render () {
     let { comments, truncate, expand } = this.props
-    const { dispatch, post, currentUser } = this.context
+    const { dispatch, post, currentUser, community } = this.context
+    let communitySlug = community ? community.slug : ''
```
levity: can you use lodash.get here and drop the case where you set the slug to an empty string? the util functions already deal with null values of slug, so it's a weird signal to the reader to set an empty string instead of null

src/components/Post.js
```diff
@@ -106,8 +110,9 @@ export const Header = ({ communities }, { post, community }) => {
 }
 Header.contextTypes = {post: object, community: object}
 
-const Details = ({ expanded, onExpand, tagLabel }, { post }) => {
-  let description = present(sanitize(post.description))
+const Details = ({ expanded, onExpand, tagLabel }, { post, community }) => {
+  let slug = community ? community.slug : ''
```
levity: same thing here

src/components/Post.js
```diff
@@ -53,16 +55,16 @@ class Post extends React.Component {
   }
 
   render () {
-    const { post, communities, comments, expanded, onExpand } = this.props
+    const { post, communities, comments, expanded, onExpand, community } = this.props
     const { tag, media } = post
     const image = find(media, m => m.type === 'image')
     const classes = cx('post', tag, {image, expanded})
-    const title = decode(post.name || '')
+    const title = formatPostTitle(decode(post.name || ''), (community || {}).slug)
```
levity: i recall that we intentionally decided not to make urls in titles into clickable links, in order to discourage the practice -- this changes that. @edwardwest can you confirm?

src/util/text.js
robbiecarlton: want me to move it into Post.js?

levity: yeah i guess

robbiecarlton:
@levity - good to merge this one now?

levity:
:+1:

PR #44, opened 2016-04-22
add test for quotes

robbiecarlton:


robbiecarlton:
@levity is this what you meant?
Or more tests than this?

levity:
that works

PR #45, opened 2016-04-22
All topics

robbiecarlton:


robbiecarlton:
@levity merged this, check it out when you have time

src/routes.js
```diff
@@ -152,5 +152,9 @@ export const communityOnboardingUrl = community =>
 export const commentUrl = comment =>
   `/p/${comment.post_id}#comment-${comment.id}`
 
-export const tagUrl = (name, slug) =>
-  slug ? `/c/${slug}/tag/${name}` : `/tag/${name}`
+export const tagUrl = (name, slug) => {
```
levity, line 155: i'm bikeshedding a bit here, but how about:

```
(slug ? `/c/${slug}` : '') + (name !== 'all-topics' ? `/tag/${name}` : '')
```

robbiecarlton, line 155: I thought about that, but it's getting a little unreadable for me. I can change it, your call...

PR #50, opened 2016-05-16
Final icons

robbiecarlton:


robbiecarlton:
@levity - I merged this, take a look when you get the chance.
There were some icons missing from the font, I'll add them when I get the new font.

PR #51, opened 2016-05-17
add mobile search bar

robbiecarlton:


robbiecarlton:
@levity - here's the search bar for mobile.
I haven't yet figured out how to have it begin off screen so the user has to scroll up to see it.
I investigated doing it in css and simple javascript, but drawing a blank, and having a hard time finding anything on google.
Is this something you know how to do, or have thoughts on a likely approach?

levity:
you could try using the functions in util/scrolling to start the page
scrolled down enough to obscure the search bar. but it's probably best to
focus on finishing the icon stuff first.

robbiecarlton:
@levity - it works! Will you take a look let me know if I can merge

PR #56, opened 2016-05-27
New ios bridge

robbiecarlton:


robbiecarlton:
@levity - take a look, and let me know if it's good to merge, but don't merge yet. I'll merge it when the new version of the app is ready in the app store

src/client/util.js
```diff
@@ -16,19 +17,21 @@ export function isAndroidApp () {
...
+    var randomName = () => {
+      return randomstring.generate({length: 10, charset: 'alphabetic'})
+    }
+
+    if (callback) {
+      const callbackIdentifier = randomName()
+      if (!window.iOSCallbacks) {
+        window.iOSCallbacks = {}
+      }
+      window.iOSCallbacks[callbackIdentifier] = callback
```
robbiecarlton: adding this to the global namespace is a little funky, but it was the only way I could figure out to make the callback accessible in wkwebview (and it seems to be the way wkwebview handles JS in general, see below `window.webkit...`)

src/client/util.js
```diff
@@ -16,19 +17,21 @@ export function isAndroidApp () {
...
-  if (window.WebViewJavascriptBridge) {
-    callback(window.WebViewJavascriptBridge)
-  } else {
-    document.addEventListener('WebViewJavascriptBridgeReady', function () {
-      callback(window.WebViewJavascriptBridge)
-    }, false)
+export function calliOSBridge (message, callback) {
+  if (isiOSApp()) {
+    var randomName = () => {
+      return randomstring.generate({length: 10, charset: 'alphabetic'})
```
levity: just use `crypto.randomBytes(2).toString('hex')` for this (increment 2 for a longer string)

src/client/util.js
```diff
@@ -16,19 +17,21 @@ export function isAndroidApp () {
...
- */
-export function connectWebViewBridge (callback) {
-  if (window.WebViewJavascriptBridge) {
-    callback(window.WebViewJavascriptBridge)
-  } else {
-    document.addEventListener('WebViewJavascriptBridgeReady', function () {
-      callback(window.WebViewJavascriptBridge)
-    }, false)
+export function calliOSBridge (message, callback) {
+  if (isiOSApp()) {
```
levity: can you just short-circuit the method here with `if (!isIOSApp()) return` to reduce indentation?

src/client/util.js
```diff
@@ -16,19 +17,21 @@ export function isAndroidApp () {
...
- * }
- */
-export function connectWebViewBridge (callback) {
-  if (window.WebViewJavascriptBridge) {
-    callback(window.WebViewJavascriptBridge)
-  } else {
-    document.addEventListener('WebViewJavascriptBridgeReady', function () {
-      callback(window.WebViewJavascriptBridge)
-    }, false)
+export function calliOSBridge (message, callback) {
```
levity, line 20: can you just call this `callMobileBridge` instead? then we can aspire to a single interface from the web app that gets handled differently behind the scenes for different platforms

which makes me wonder... dare i ask, how does this impact the Android situation, if at all?

src/client/util.js
levity: then you can remove the `randomstring` dependency

src/client/util.js
robbiecarlton, line 20: I vote to keep it as it is.
I think we'll always want to have two different functions, one for Android and one for iOS, partly because Android has way richer js bridge functionality which makes the code in the android app simpler, but would be hard/defeat the point to reduce it to the same interface as the iOS bridge, and partly because there are times when we may want to call one and not the other. 

these changes won't impact the android situation. The Android bridge is the same and being called in the same way for Filepicker. 
There are likely some little bugs in the android app that need ironing out with the new release.

levity, line 20: OK, I'll defer to you here.

PR #57, opened 2016-06-01
left nav updates tag list when post is saved

robbiecarlton:


robbiecarlton:
@levity - merged this, check it out when you get the chance

PR #58, opened 2016-06-01
fix post editor update issue on mobile

robbiecarlton:


robbiecarlton:
@levity - merged this, take a look when you get a moment. I ended up using tagSelector when available, as it was less jarring than switching focus to title.

PR #60, opened 2016-06-02
Topic notifications

robbiecarlton:


src/containers/App.js
```diff
@@ -65,11 +65,40 @@ export default class App extends React.Component {
     return pick(this.props, 'dispatch', 'currentUser', 'isMobile')
   }
 
+  setLeftNavRefreshInterval (community) {
```
robbiecarlton: I'm doing this in App so that LeftNav can remain a stateless component.

robbiecarlton:
@levity - good to merge?

src/containers/App.js
levity: hmmmm --- can you integrate this into fetchLiveStatus instead, to minimize the number of request-response roundtrips?

src/containers/tag/TagPosts.js
```diff
@@ -66,11 +66,13 @@ class TagPosts extends React.Component {
 }
 
 export default compose(
-  prefetch(({ dispatch, params: { tagName, id }, query }) =>
-    dispatch(fetchTag(tagName, id))
+  prefetch(({ dispatch, params: { tagName, id }, query }) => {
+    dispatch(resetNewPostCount(tagName, id))
```
levity: can you just do `if (id) dispatch(resetNewPostCount(tagname, id))` here and drop the `NO_ACTION` stuff in the action?

src/server/appHandler.js
```diff
@@ -69,7 +69,7 @@ export default function (req, res) {
   .catch(err => {
     res.errors = [err]
     res.setHeader('Content-Type', 'text/plain')
-    const state = parse(req.url, true).query.verboseErrorPage
+    const state = parse(req.originalUrl, true).query.verboseErrorPage
```
levity, line 72: what's this for?

robbiecarlton, line 72: I was getting an error in this branch (in addition to the error that is being caught by this branch) because the req object didn't have a url key. It did have an originalUrl key, which fixed the error.

robbiecarlton:
@levity - made those changes, good to mere now?

src/components/LiveStatusPoller.js
```diff
@@ -1,20 +1,41 @@
...
+      clearInterval(this.pollInterval)
+    }
+
+    this.pollInterval = setInterval(() => dispatch(fetchLiveStatus(communityId)),
+      6 * 1000)
+  }
+
+  componentDidMount () {
+    let { dispatch, communityId } = this.props
+    dispatch(fetchLiveStatus(communityId))
```
levity: can we defer this initial fetch for, say, 10 seconds, just to reduce the amount of work that happens during mounting?

levity:
looks good otherwise

PR #63, opened 2016-06-14
Hashtag popover

robbiecarlton:


robbiecarlton:
@levity - good to merge?

src/components/TagPopover.js
```diff
@@ -0,0 +1,120 @@
...
+    if (isEmpty(nextProps.tagName)) return
+
+    // loading
+    if (nextProps.tagName !== tagName || nextProps.slug !== slug) {
+      dispatch(fetchTagSummary(nextProps.tagName, nextProps.slug))
+    }
+
+    // content has loaded
+    if (nextProps.description) {
+      this.hideTimeoutId = setTimeout(() => this.hide(), 3000)
```
robbiecarlton: I used this instead of the document.onMouseover listener, because the latter was not working. It would often be triggered when moving from the tag link to the popover, and so hide the popover before you could get to it.

css/index.less
```diff
@@ -863,6 +863,85 @@ ul.tags {
...
+  .footer {
+    .meta-text;
+    background-color: @faint-gray-bg;
+    padding: 15px 15px;
+    a {
+      color: @damian-green;
+    }
+  }
+}
+.popover.below:after, .popover.below:before {
```
levity: can you nest these in the above section?

css/index.less
```diff
@@ -863,6 +863,85 @@ ul.tags {
...
+  left: 50%;
+  border: solid transparent;
+  content: " ";
+  height: 0;
+  width: 0;
+  position: absolute;
+  pointer-events: none;
+}
+.popover.below:after {
+	border-bottom-color: white;
```
levity: indentation

src/reducers/tags.js
```diff
@@ -76,14 +76,24 @@ export const tagsByCommunity = (state = {}, action) => {
       return mergeLeftNavTags(state, payload.left_nav_tags, get(meta, 'slug'))
     case FOLLOW_TAG_PENDING:
       oldCommunityTags = state[meta.id] || {}
-      var oldTag = oldCommunityTags[meta.tagName]
```
levity, line 79: this and `oldCommunityTags` should be declared outside of any switch case

src/actions/index.js
```diff
@@ -646,6 +649,22 @@ export function editTagDescription (tag, description) {
...
+
+export function hideTagPopover () {
+  return {type: HIDE_TAG_POPOVER}
+}
+
+export function fetchTagSummary (tagName, id) {
+  return {
+    type: FETCH_TAG_SUMMARY,
+    payload: {api: true, path: `/noo/community/${id}/tag/${tagName}/summary`},
+    meta: {tagName, id}
```
levity, line 664: you should use caching for this action so if the mouseover is used more than once it's snappy

src/components/ClickCatcher.js
```diff
@@ -17,11 +19,20 @@ const ClickCatcher = ({ tag, ...props }, { dispatch }) => {
       node.setAttribute('target', '_blank')
     }
   }
+
+  const handleMouseOver = event => {
+    var node = event.target
+    if (node.nodeName.toLowerCase() === 'a' && node.getAttribute('class') === 'hashtag') {
+      let { tagName, slug } = tagUrlComponents(node.getAttribute('href'))
+      dispatch(showTagPopover(tagName, slug, positionInViewport(node), node.offsetWidth))
```
levity: why did you use `positionInViewport`?

levity: should work the same with `position` if you drop `window.pageYOffset` in the component

src/components/TagPopover.js
```diff
@@ -0,0 +1,120 @@
...
+    const popoverWidth = 290
+    let left = position.x - 25 + (anchorWidth / 2) - (popoverWidth / 2)
+    if (left < 0) {
+      left = 0
+    } else if (left > document.documentElement.clientWidth - (popoverWidth + 30)) {
+      left = document.documentElement.clientWidth - (popoverWidth + 30)
+    }
+    const above = position.y > popoverHeight + 70
+    let outerPosition = {left}
+    outerPosition.top = window.pageYOffset + position.y + 15
```
levity: you can move this assignment into the initial declaration of `outerPosition`

src/components/TagPopover.js
levity: that could have been addressed by adjusting the margin & positioning of the popover so that it actually covers the entire link, but ¯_(ツ)_/¯

src/actions/index.js
robbiecarlton, line 664: It actually is cached, in that, the old data is already in the store and displays instantly, and then this is called and if there's any change to the data it's updated when this returns

src/components/ClickCatcher.js
robbiecarlton: ah, yes, this makes sense. There was some reason I switched when exploring solutions, but don't need it now.

src/actions/index.js
levity, line 664: ideally we'd just skip subsequent API calls--i don't think the data's going to change frequently enough to justify reloading it within a single session

src/components/TagPopover.js
levity: does this mean that if you mouse over a link but don't move the mouse onto the popover, it will disappear?

css/index.less
robbiecarlton: weird. Indentation is not like that in the file

src/components/ClickCatcher.js
robbiecarlton: actually, I do need it now. I need position in viewport to decide whether to display the popover above or below the link

src/components/TagPopover.js
robbiecarlton: after 3 seconds, yes

On Tue, Jun 14, 2016 at 3:45 PM, Lawrence Wang notifications@github.com
wrote:

> In src/components/TagPopover.js
> https://github.com/Hylozoic/hylo-redux/pull/63#discussion_r67071801:
> 
> > -  }
> >   +
> > -  componentWillReceiveProps (nextProps) {
> > -    let { dispatch, slug, tagName } = this.props
> >   +
> > -    // hidden
> > -    if (isEmpty(nextProps.tagName)) return
> >   +
> > -    // loading
> > -    if (nextProps.tagName !== tagName || nextProps.slug !== slug) {
> > -      dispatch(fetchTagSummary(nextProps.tagName, nextProps.slug))
> > -    }
> >   +
> > -    // content has loaded
> > -    if (nextProps.description) {
> > -      this.hideTimeoutId = setTimeout(() => this.hide(), 3000)
> 
> does this mean that if you mouse over a link but don't move the mouse onto
> the popover, it will disappear?
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> https://github.com/Hylozoic/hylo-redux/pull/63/files/d465fd1fd1ffc2a58ffde8eed20f2bdaefdf77c1#r67071801,
> or mute the thread
> https://github.com/notifications/unsubscribe/AA2Y9N58cCUbZ8VM2ms6ZnOTSsuIoTJUks5qLy8ggaJpZM4I1ybD
> .

## 

Robbie Carlton

719.358.3482
robbie@authenticmanprogram.com

src/reducers/tags.js
```diff
@@ -60,9 +60,12 @@ export const tagsByCommunity = (state = {}, action) => {
   let { type, payload, meta, error } = action
   if (error) return state
 
+  let oldCommunityTags
```
levity: it's ok to use `let x, y` when they're not being assigned values

robbiecarlton:
@levity - how's this now?
Only thing I didn't do from your list was nest the popover callout css inside the popover class. Couldn't figure out how to do this without breaking it, and I wanted to merge this because Ed wants this live.
I can go back and figure out how to clean up the css when my brain is fresher. Does that work for you?

levity:
:ok:

levity:
it should just be `&.below:before, &.below:after`

robbiecarlton:
yup that did it.

On Tue, Jun 14, 2016 at 5:32 PM, Lawrence Wang notifications@github.com
wrote:

> it should just be &.below:before, &.below:after
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> https://github.com/Hylozoic/hylo-redux/pull/63#issuecomment-226056689,
> or mute the thread
> https://github.com/notifications/unsubscribe/AA2Y9BXfVaIo7EHcPCbLdQWrArhr3LURks5qL0g3gaJpZM4I1ybD
> .

## 

Robbie Carlton

719.358.3482
robbie@authenticmanprogram.com

PR #66, opened 2016-06-20
Post topic dropdown

robbiecarlton:


robbiecarlton:
@levity - good to merge?

css/post-editor.less
```diff
@@ -371,6 +382,9 @@
       padding: 0;
       border: 0;
       background-color: transparent;
+      &.tag {
```
levity: what's this for?

src/reducers/postEdits.js
```diff
@@ -122,7 +125,13 @@ export const tagDescriptionEdits = (state = {}, action) => {
...
     return {}
   } else if (type === EDIT_TAG_DESCRIPTION) {
-    return {...state, [payload.tag]: payload.description}
+    if (state.creating === true) {
+      return {creating: true, [payload.tag]: payload.description}
+    } else {
+      return {...state, [payload.tag]: payload.description}
+    }
+  } else if (type === CREATE_TAG_IN_POST_EDITOR) {
+    return {creating: true, '': ''}
```
levity: rather than doing this really weird-looking `{'': ''}` thing, can you just have code in `TagDescriptionEditor` that knows to set a blank tag if none is found?

levity:
tests please

src/components/PostEditor.js
```diff
@@ -302,12 +308,13 @@ export class PostEditor extends React.Component {
       editingTagDescriptions
     } = this.props
     const { currentUser } = this.context
-    const selectableTags = compact([this.props.tag].concat(specialTags))
+    const selectableTags = compact([this.props.tag, post.tag].concat(specialTags))
```
levity: need to remove duplicates and blank values here

src/reducers/postEdits.js
```diff
@@ -122,7 +125,13 @@ export const tagDescriptionEdits = (state = {}, action) => {
   } else if (includes([START_POST_EDIT, CANCEL_POST_EDIT, CREATE_COMMENT], type)) {
     return {}
   } else if (type === EDIT_TAG_DESCRIPTION) {
-    return {...state, [payload.tag]: payload.description}
+    if (state.creating === true) {
```
levity: doesn't the if-clause here return the same thing that would be returned if the if-clause wasn't present?

css/post-editor.less
levity: oh never mind

src/components/TagDescriptionEditor.js
```diff
@@ -43,7 +75,9 @@ export default class TagDescriptionEditor extends React.Component {
           </div>
         </div>)}
         <div className='footer'>
-          <button onClick={() => onSave(tags)} className='ok'>Create</button>
+          {creating
+          ? <button onClick={createTag} className='ok'>Create</button>
```
levity: indent the ? and :

src/components/TagDescriptionEditor.js
```diff
@@ -34,7 +63,10 @@ export default class TagDescriptionEditor extends React.Component {
         {map(tags, (description, tag) => <div key={tag} className='tag-group'>
           <div className='topic'>
             <label>Topic</label>
-            <span>#{tag}</span>
+            {creating
+            ? <span>#&nbsp;<input type='text' defaultValue={tag} className='tag'
```
levity: move the <input> to its own line -- there's an implicit style rule that we've been consistently following, which is that if a tag is on the same line as its parent's start tag, then the parent's end tag must also be on that same line

levity: the input should probably have a placeholder value like "topic name"

src/reducers/postEdits.js
levity: better to set up a different reducer for the creating toggle rather than shoehorn it in here. then you can drop the code in the component that looks for a special value and omits it

src/components/TagDescriptionEditor.js
levity: actually in this case the only difference in the two values is the onClick handler so you might as well shrink the conditional expression to change just that

src/components/PostEditor.js
```diff
@@ -342,9 +349,10 @@ export class PostEditor extends React.Component {
         </button>
       }>
         {selectableTags.map(t => <li key={t}>
-          <a onClick={() => selectTag(t)}>#{t}</a>
+          <a onClick={() => selectTag(t)}># {t}</a>
```
levity: i know this is in the mockup but i don't think it's a good idea

src/components/PostEditor.js
levity: actually, shouldn't this be `tag` as declared in the next line from `postEdit`, not `post`?

src/reducers/postEdits.js
robbiecarlton: no, because if `creating === true`, then payload.tag keeps changing, so if you merge `{[payload.tag]: payload.description}` with previous state, you end up with

```
{
  t: 'description',
  ta: 'description',
  tag: 'description',
  tagn: 'description',
  tagna: 'description'
}
```

this is also why I elected to store creating: true in this reducer rather than giving it it's own, because this reducer needs to know which mode we're in.
I'll think about if there's another way to do it, and if you see one please let me know, I'd **love** to not have this `creating === true` stuff.

robbiecarlton:
@levity - good to merge?

test/reducers/postEdits.test.js
```diff
@@ -169,3 +172,80 @@ describe('postEdits', () => {
...
+      const state = {}
+      const expectedState = {
+        [tag1]: '',
+        [tag2]: ''
+      }
+      expect(tagDescriptionEdits(state, action)).to.deep.equal(expectedState)
+    })
+  })
+
+  describe('on EDIT_TAG_DESCRIPTION it updates the correct tag', () => {
```
levity: the description for describe shouldn't have a clause in it

src/components/TagDescriptionEditor.js
```diff
@@ -1,34 +1,43 @@
...
     updatePostTag: func,
-    dispatch: func
+    dispatch: func,
+    creating: bool
+  }
+
+  componentWillReceiveProps (nextProps) {
+    let { tags, creating } = this.props
+    if (creating && keys(nextProps.tags)[0] !== keys(tags)[0]) {
+      console.log('didnt equal')
```
levity: remove console.log

src/components/TagDescriptionEditor.js
```diff
@@ -1,27 +1,64 @@
...
+    saveTagDescriptions: func,
+    updatePostTag: func,
+    dispatch: func,
+    creating: bool
+  }
+
+  componentWillReceiveProps (nextProps) {
+    let { tags, creating } = this.props
+    if (creating && keys(nextProps.tags)[0] !== keys(tags)[0]) {
+      this.refs.tag.focus()
```
levity: could this just be in `componentDidMount`?

robbiecarlton: actually, it's redundant now

levity: even better :)

PR #68, opened 2016-06-25
Save Partial Comments

robbiecarlton:
@levity - would you check out this approach to writing a test for 'Save partially edited comment to app state as you type'

(it doesn't currently work because of some interaction with tinyMCE, but just the general approach).
Is there something I'm missing that could make this simpler? Or is this kind of test to brittle anyway?
It's taking me a long time to get it right, and I want a sanity check.


levity:
I've been really wary of writing tests around TinyMCE because I have a hard time imagining it working in a meaningful way without a real browser environment. I think if we want to test these kinds of interactions thoroughly (and we do!) we should set up a PhantomJS or other test environment that uses an actual headless browser.

harlantwood:
http://www.nightmarejs.org was the leading edge of this tech when I last researched it about half a year ago.  Still painful, browser testing headless or otherwise always seems to be, in my experience.  Still a great idea to do it though.

levity:
Thanks for the suggestion, Harlan!

harlantwood:
Sure thing!

robbiecarlton:
@levity - good to merge?

test/client/components/CommentForm.test.js
```diff
@@ -0,0 +1,71 @@
...
+            '<p>A long enough description that the show more button is displayed.</p>' +
+            '<p>A long enough description that the show more button is displayed.</p>'
+        }
+      },
+      totalPostsByQuery: {'subject=community&id=foo': 1},
+      postsByQuery: {'subject=community&id=foo': [1]},
+      pending: {}
+    }
+
+    it.skip('retains comment data after blur', () => {
```
robbiecarlton, line 47: I left this in because it took me a long time to write ;) and it might be salvageable once we set up the browser testing

levity:
:+1:

PR #71, opened 2016-06-29
Simplified left nav tags

levity:


levity:
Goes with Hylozoic/hylo-node#140 -- @robbiecarlton, pls review

robbiecarlton:
👍

PR #73, opened 2016-07-01
Remember community last visited and start with that upon return

robbiecarlton:


robbiecarlton:
@levity - take a look when you get the chance.

I couldn't figure out how to write tests for the stuff happening in the prefetch of AllPosts. If you have any suggestions, let me know.

src/containers/home.js
```diff
@@ -5,20 +5,45 @@ import { connect } from 'react-redux'
...
+  }
+})
 
 const wrapComponent = (subject, ...args) => compose(
-  prefetch(({ dispatch, params, currentUser: { id }, query }) => {
-    dispatch(setCurrentCommunityId('all'))
+  prefetch(({ dispatch, params, currentUser: { id, settings: { currentCommunityId } }, query, path }) => {
+    if (query.rd && currentCommunityId) {
+      if (currentCommunityId !== 'all') {
+        // we only have the id, so have to fetch to get the slug
```
levity: we can avoid a network round-trip here by searching within the current user's memberships for the community with the id

levity: also, if you want to test this, pull the prefetch function out and export it so that you can call it from a test. then just give it mock arguments and test that it dispatches `navigate` with the right url

src/actions/util.js
```diff
@@ -7,3 +8,10 @@ export const findError = (errors, type, bucket, id) => {
 
   return get(find([errors[type]], match), 'payload.response')
 }
+
+export const setCurrentCommunityIdLocalAndRemote = (dispatch, communityId, userId) => {
+  if (!communityId) return
+  const settings = {currentCommunityId: communityId}
+  userId && dispatch(updateUserSettings(userId, {settings}))
```
levity, line 15: btw, i think we have a kind-of-serious bug on the server side where updating one value in settings causes all the other existing values to be dropped. can you fix that and write a test for it?

levity, line 15: let's rename this to `saveCurrentCommunityId` or something for brevity's sake

robbiecarlton, line 15: see https://github.com/Hylozoic/hylo-node/pull/142

robbiecarlton:
@levity - good?

src/util/auth.js
```diff
@@ -53,8 +56,12 @@ export function setupPopupCallback (name, dispatch, errorAction) {
           .then(action => {
             if (action.error) return
             let params = qs.parse(window.location.search.replace(/^\?/, ''))
-            let next = params.next || `/u/${action.payload.id}`
-            dispatch(navigate(next))
+
+            const currentUser = action.payload
+            const currentCommunityId = get(currentUser, 'settings.currentCommunityId')
+            const community = getCommunity(currentUser, currentCommunityId)
+            next = params.next || community ? communityUrl(community) : `/u/${action.payload.id}`
```
levity: need parens:

``` js
sails> 4 || true ? 1 : 2
1
sails> 4 || (true ? 1 : 2)
4
```

src/containers/home.js
```diff
@@ -5,22 +5,48 @@ import { connect } from 'react-redux'
...
+  }
+})
+
+export const prefetchForWrapped = subject => ({ dispatch, params, currentUser, query, path }) => {
+  const { id, settings } = currentUser
+  const { currentCommunityId } = settings
+  if (query.rd && currentCommunityId) {
+    if (currentCommunityId !== 'all') {
+      const community = getCommunity(currentUser, currentCommunityId)
+      community && dispatch(navigate(communityUrl(community)))
```
levity: you should return here to avoid running the lines below

src/containers/home.js
```diff
@@ -5,22 +5,48 @@ import { connect } from 'react-redux'
...
+      <ConnectedPostList {...{subject, id, query}}/>
+    </CoverImagePage>
+  }
+})
+
+export const prefetchForWrapped = subject => ({ dispatch, params, currentUser, query, path }) => {
+  const { id, settings } = currentUser
+  const { currentCommunityId } = settings
+  if (query.rd && currentCommunityId) {
+    if (currentCommunityId !== 'all') {
```
levity: collapse these two ifs

levity:
:+1:

PR #75, opened 2016-07-07
Pinned posts

robbiecarlton:
Goes with https://github.com/Hylozoic/hylo-node/pull/144/

### DIFF
--- a/src/reducers/util.js
+++ b/src/reducers/util.js
@@ -1,6 +1,6 @@
 import {
-  cloneDeep, filter, get, includes, isArray, isEqual, isObject, mergeWith, set, some,
-  transform, uniqBy, uniq
+  cloneDeep, filter, get, includes, isArray, isEqual, mergeWith, set, some,
+  transform, uniqBy
 } from 'lodash'
 import { CLEAR_CACHE } from '../actions'
 
@@ -35,10 +35,6 @@ export function updateMedia (obj, type, url) {
   return {...obj, media}
 }
 
-const isSimpleArray = array => isArray(array) && (array.length === 0 || !isObject(array[0]))
-
-const mergeArray = (dest, src) => uniq(dest.concat(src))
-
 // update state with a set of items. items that already exist in the state get
 // new properties, update the values of existing properties, and do not lose any
 // existing properties that aren't contained in their new counterpart.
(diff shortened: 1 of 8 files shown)


robbiecarlton:
@levity still fixing a couple of tests. But I want to validate the approach with you. Specifically, would you take a look at how I've handled storing the pinned info on posts? Basically, when fetchPosts is called for a community, I return a bool 'pinned' field for each post, that refers to that community. Then in the store I translate this into a per post array called 'pins' of community ids where the post is pinned.

It feels a little hacky, and not sure how nicely it plays with caching yet, but I couldn't find a cleaner solution. Thoughts?

src/reducers/util.js
```diff
@@ -35,6 +35,10 @@ export function updateMedia (obj, type, url) {
   return {...obj, media}
 }
 
+const isSimpleArray = array => isArray(array) && (array.length === 0 || !isObject(array[0]))
+
+const mergeArray = (dest, src) => uniq(dest.concat(src))
```
levity: the logic behind using `uniq` here seems way too specific to pinning to fold into `mergeList`. it seems like it would lead to subtle & difficult-to-diagnose bugs. The underlying assumption you're enforcing here is that any time you have an array of numbers or strings, they are ID's and so you want them to be unique -- that seems like an awkward thing to impose upon everyone who uses or might want to use `mergeList`

levity: `isObject` is also... interesting...

``` js
> isObject('foo')
false
> isObject(new String('foo'))
true
```

robbiecarlton:
@levity - good to merge?

src/reducers/util.js
```diff
@@ -44,8 +44,8 @@ export const mergeList = (state, items, key) => {
     m[id] = mergeWith({...state[id]}, x, (objValue, srcValue) => {
       // we don't want to perform the default behavior of merge when working
       // with arrays, because that concatenates them. e.g. with the list of
-      // media for a post, that creates duplicate values. so we replace the old
-      // value with the new one instead.
+      // media for a post, that creates duplicate values. so instead we
+      // replace the old value with the new one if they contain objects.
```
levity: revert to the old comment here

levity:
:+1: otherwise

PR #76, opened 2016-07-08
Remove Community Member

robbiecarlton:


src/components/PersonCards.js
```diff
@@ -5,14 +5,15 @@ import Masonry from 'react-masonry-component'
...
-const PersonCards = ({ people, menus, slug, onSkillClick }) => {
+const PersonCards = ({ people, menus, slug, onSkillClick, removeMember }) => {
   return <div className='person-cards masonry-container'>
     <Masonry options={{transitionDuration: 0}}>
       {people.map(person => <div key={person.id} className='masonry-item-wrapper'>
         <PersonCard person={person}
           menu={menus && menus[person.id]}
           slug={slug}
           onSkillClick={onSkillClick}
+          removeMember={removeMember}
```
robbiecarlton, line 15: I didn't use the `menus` prop, but it doesn't make sense to have them both there. I can rewrite to use the menus prop, or just remove it. What do you think?

robbiecarlton:
@levity - please take a look when you get the chance

robbiecarlton:
@levity - bump

src/actions/index.js
```diff
@@ -376,6 +378,14 @@ export function removeCommunityModerator (community, moderatorId, prevProps) {
   }
 }
 
+export function removeCommunityMember (community, userId, cacheId, prevProps) {
+  return {
+    type: REMOVE_COMMUNITY_MEMBER,
+    payload: {api: true, path: `/noo/community/${community.id}/member/${userId}`, method: 'DELETE'},
+    meta: {slug: community.slug, userId, cacheId, prevProps}
```
levity: it feels wrong to me to have this action depend on cacheId and prevProps -- it's really the reducer that should figure out which caches have this member in them and remove the member's id.

for example, if i load the members page, and the person i'm looking to remove is somewhere on the page, but i don't see it, i might search for their name. so now in cache we have `subject=community&id=foo` and `subject=community&id=foo&search=tim`. if i call `removeCommunityMember` now, it should remove it from both caches.

the presence of `prevProps` complicates things... i think it's too coupled now to do optimistic updating this way. i'm not sure how else to do it (just spent a few minutes thinking about it, didn't come up with a clear answer), but perhaps we should just do things non-optimistically for now until we have time to come up with a better solution.

src/components/PersonCards.js
levity, line 15: we don't seem to be setting `menus` anymore (it was part of the first Redux implementation, pre-Damian), so you can just drop it and stick with what you've done.

src/actions/index.js
levity: the answer probably has to do with middleware -- say, setting `payload.optimistic` triggers middleware that saves a snapshot of the app state, and sets up a promise chain that reverts to the snapshot if the final API action returns with `error` set

robbiecarlton: Yes, agreed prevProps and cacheId is fishy here.
Middleware idea is smart. I think it would have to save a diff, because other actions can potentially fire in between.
So for now should I just remove the optimistic stuff here?

levity: yeah -- unless you want to take a stab at the middleware

robbiecarlton: ok, I'll give it a shot, with a time limit.

On Tue, Jul 12, 2016 at 1:16 PM, Lawrence Wang notifications@github.com
wrote:

> In src/actions/index.js
> https://github.com/Hylozoic/hylo-redux/pull/76#discussion_r70513553:
> 
> > @@ -376,6 +378,14 @@ export function removeCommunityModerator (community, moderatorId, prevProps) {
> >    }
> >  }
> > 
> > +export function removeCommunityMember (community, userId, cacheId, prevProps) {
> > -  return {
> > -    type: REMOVE_COMMUNITY_MEMBER,
> > -    payload: {api: true, path: `/noo/community/${community.id}/member/${userId}`, method: 'DELETE'},
> > -    meta: {slug: community.slug, userId, cacheId, prevProps}
> 
> yeah -- unless you want to take a stab at the middleware
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> https://github.com/Hylozoic/hylo-redux/pull/76/files/17410bee3d125958c8ae4e94b3dabcef767825a6#r70513553,
> or mute the thread
> https://github.com/notifications/unsubscribe/AA2Y9CyCUaK94WRO41_VQeDW8txZoz5oks5qU_YjgaJpZM4JIaXw
> .

## 

Robbie Carlton

719.358.3482
robbie@authenticmanprogram.com

src/reducers/index.js
```diff
@@ -75,7 +75,7 @@ import {
   VALIDATE_NETWORK_ATTRIBUTE_PENDING
 } from '../actions'
 
-export default combineReducers({
+export default combineReducers(injectSetState({
```
levity: the output of combineReducers here is also just one giant reducer, so rather than modifying each sub-reducer, you can do:

``` js
const combinedReducers = combineReducers(...)
const handleSetState = (state = {}, { type, payload }) => type === SET_STATE ? payload : state
export default composeReducers(combinedReducers, handleSetState)
```

robbiecarlton:
@levity - take a look when you get the chance

test/reducers/util.test.js
```diff
@@ -10,3 +12,35 @@ describe('toggleIncludes', () => {
...
+    if (action.type === 'ADD_C') {
+      return {...state, c: action.payload}
+    }
+    return state
+  }
+  const stringReducer = (state = 'default', action) => {
+    return state
+  }
+
+  it('injects a SET_STATE clause in each reducer', () => {
```
levity: this name needs to be updated

levity:
other than the test name, :+1:

PR #78, opened 2016-07-16
Link previews

levity:


levity:
@robbiecarlton pls review

robbiecarlton:
I took a look. I couldn't see where you are storing the data from FETCH_LINK_PREVIEW. Am I missing something?

levity:
Oh, yeah, I'm not, at least in the usual way. The post editor is taking the
value from the action directly via .then and storing it the same way as
other editor properties. I guess I could just move that bit into the
postEdits reducer.

levity:
What I do like about this way though is how decoupled and "pure" the fetchLinkPreview action is. It just has a URL, no extra arguments that tie it to a post-editing context. If/when we want to do this for comments as well, it's ready for reuse.

robbiecarlton:
ah, I see it. Makes sense. My default is to do everything the redux way, but maybe this keeps it simpler.

levity:
:+1: ?

robbiecarlton:
👍

PR #81, opened 2016-07-20
Update Favicon and Page Title for Community

robbiecarlton:


robbiecarlton:
@levity - I couldn't figure out a way to test this as it's all happening in the browser, not the rendered document. If you can see a way, short of setting up nightmarejs, let me know.
Otherwise, good to merge?

src/containers/SinglePost.js
```diff
@@ -141,7 +142,7 @@ const setupPage = (store, id, query, action) => {
   }
 
   return Promise.all([
-    saveCurrentCommunityId(dispatch, communityId, userId),
+    saveCurrentCommunity(dispatch, community, userId),
```
levity: trying to understand the change from `saveCurrentCommunityId` to `saveCurrentCommunity` -- is it necessary? this appears to be the only case where passing community rather than id would make a difference, but in this case (where no community is found for some reason) we'd want to display the default favicon, no?

src/components/PageTitleController.js
```diff
@@ -1,28 +1,66 @@
...
+    if (!this.originalFaviconUrl) {
+      this.originalFaviconUrl = oldLink.href
+    }
+    link.id = 'favicon'
+    link.rel = 'shortcut icon'
+    link.href = faviconUrl
+    if (oldLink) {
+      document.head.removeChild(oldLink)
+    }
+    document.head.appendChild(link)
```
levity: does editing the value of the existing link tag not work?

src/containers/SinglePost.js
robbiecarlton: This was left over from my first approach, where I needed the community name in this function. Fixed now.

src/components/PageTitleController.js
robbiecarlton: Yes it does. Updated.

levity:
ah, now there's a nice clean diff :)
:+1:

PR #85, opened 2016-07-27
Nav updates

robbiecarlton:


robbiecarlton:
@levity - good to merge?

robbiecarlton:
@levity - btw, couldn't test the full width version because I don't have a wide enough screen. Don't anticipate any problems, but could you double check?

src/containers/App.js
```diff
@@ -35,7 +36,7 @@ const makeNavLinks = (currentUser, community) => {
     {url: network && `/n/${network.slug}`, icon: 'merkaba', label: 'Network'},
     {url: slug && url('about'), icon: 'Help', label: 'About'},
     {url: canInvite(currentUser, community) && url('invite'), icon: 'Mail', label: 'Invite'},
-    {url: canModerate(currentUser, community) && url('settings'), icon: 'Settings', label: 'Settings'}
+    {url: canModerate(currentUser, community) && url('settings'), icon: 'Settings', label: 'Admin Settings'}
```
levity: we settled on "Community Settings"

src/containers/App.js
```diff
@@ -158,12 +158,15 @@ export default class App extends React.Component {
         path={path}
         search={doSearch}
         opened={leftNavIsOpen}
-        networkNavAnimation={moveWithMenu}
-        networkNavEasing={leftNavEasing}
         />}
 
       <VelocityComponent animation={moveWithMenu} easing={leftNavEasing}>
         <div id='main'>
+          {currentUser && !isMobile && networkCommunities &&
```
levity: this is a lot of clauses; can we just wrap them up into a `showNetworkNav` const?

levity:
I'm noticing that the extra menu item fits better when the menu is closed than when it's open. We should be factoring that into deciding how many links to show up there (which we can do with CSS since the main div gets the `leftNavIsOpen` class).

I would reduce the max-width of the community name from 400px to 240px also

robbiecarlton:
@levity - I have to stop for today, didn't get a chance to finish the variable width depending on leftNavIsOpen. 
If this looks good otherwise, can you merge it, to fix the profile bug ed was talking about.
I'll finish the variable width stuff tomorrow

levity:
sounds good

PR #86, opened 2016-07-28
Run the name through sanitization to prevent name XSS issue

aaron-m-edwards:
Commit fixes a XSS injection site where your name can contain tags which are not stripped out. The only location found that displays the tags without escaping or sanitisation that was found was when displaying comments

screen shot 2016-07-28 at 4 35 18 pm

---

screen shot 2016-07-28 at 4 35 42 pm


levity:
Ah, good catch. Thanks!

PR #87, opened 2016-07-28
adjust main menu width when leftnav is open

robbiecarlton:


css/nav.less
```diff
@@ -488,6 +488,24 @@ nav {
   }
 }
 
+.leftNavIsOpen #topNav .main-menu {
+  @media (max-width: (@main-menu-hide-breakpoint + @left-nav-width - 1)) {
+    display: none;
```
levity, line 493: this feels wrong to me -- it makes the menu jump between the left nav and the top nav when you toggle the left nav. will things still fit if you just remove this rule?

robbiecarlton, line 493: I could adjust the widths so that things could still fit, what this would mean is just setting  the break point higher, so not using the extra room.

PR #91, opened 2016-08-03
more items in mention autocomplete list, and shorten dropdown

robbiecarlton:


robbiecarlton:
@levity - i merged this, check it out when you can

css/index.less
```diff
@@ -356,6 +356,7 @@ ul.tags {
   }
   > .dropdown {
     overflow: visible;
+    max-height: 31vh;
```
levity, line 359: as far as i can tell, this doesn't have any effect on the size of the `ul.dropdown-menu` child, whose max-height is set in dropdown.less

robbiecarlton, line 359: you're right. I moved it into dropdown-menu, but kept it in index.less rich-text-editor, because it's specific to the rich-text-editor

PR #93, opened 2016-08-04
Re-add slack integration section

Connoropolous:
- [x] only unhandled thing is displaying an error... how?

I don't need the error to be specific, just to let them know there was one. There will be a value in the querystring `slackerror=true`


levity:
You can use the `notify` action to trigger Notifier to show a message in
the lower right

Connoropolous:
@levity I am struggling to understand where in the code I should put something that looks at the location, gets a param off the query string, and then shows the error message using notify. could you advise?

levity:
Sure! Because the CommunitySettings component is listed in a `<Route>` in routes.js, it gets props from react-router, including `location`. So from anywhere that you can reference `this.props`, you can get the query parameter you're looking for at `this.props.location.query.slackerror`. 

Now as for the question of when/where to trigger the error: On second thought, I wouldn't use `notify` in this case, because it causes a transient notification that fades after 5 seconds. Since we're reloading the entire page in this case, it's probably better to have a persistent error notification. So in `render`, I would add a `<div className='alert alert-danger'>` to the Slack section of CommunitySettings that only gets shown when `this.props.location.query.slackerror` is set.

Connoropolous:
super. will do!

Connoropolous:
@levity this is ready now

PR #94, opened 2016-08-05
Tooltips

robbiecarlton:


robbiecarlton:
screen shot 2016-08-05 at 12 03 39 pm
screen shot 2016-08-05 at 12 03 56 pm
screen shot 2016-08-05 at 12 05 21 pm
screen shot 2016-08-05 at 12 05 43 pm
screen shot 2016-08-05 at 12 06 01 pm

robbiecarlton:
@levity - could you take a look at this.
Both the code, but also the tests are failing running `cover`
I get the same error running `cover` locally on the master branch. Not sure what's happening.

levity:
Do you get the same behavior as in the CircleCI run? Locally, I get that stack trace, but then it goes on to finish all the tests.

levity:
I might be missing something about the need for the tooltips reducer. Why not just have the value in currentUser.settings, and then read that in @connect for the parent components to decide whether to show their tooltip or not?

robbiecarlton:
I get the same thing as you locally

robbiecarlton:
The reducer allows us to control the order that the tooltips are displayed in. When there are multiple tooltips on the page at the same time, each tooltip knows which other tooltips want to be displayed, and then can determine if it is next in the priority order. The reducer stores an id with an index for each tooltip that could potentially be displayed, and then each tooltip checks if it is the highest priority, and if so it is displayed. 
Make sense now?

robbiecarlton:
Maybe what you're saying is just put that all in user settings? We could, but as it's to do with local app state, it seems weird to store it in the db

PR #104, opened 2016-08-19
Project card tweaks

robbiecarlton:
screen shot 2016-08-19 at 1 46 05 pm
screen shot 2016-08-19 at 1 45 26 pm

Couple of difference from the design (attached below)
Specifically, don't think we have "Draft" projects right now, so I kept the tag in that place, and we don't have % funded, so didn't add that.

Original design:
project-card


robbiecarlton:
@levity - good to merge?

levity:
:+1:

PR #106, opened 2016-08-19
Style changes to comment button

robbiecarlton:


robbiecarlton:
@levity - good to merge?

I had to use onKeyDown to track the changes, because onChange is not reliable for that. Wondering if we want to call setText from onKeyDown as well, which would be another way to solve that bugs where the store doesn't get all the content from the text editor on post.

Make sense?

src/components/CommentForm.js
```diff
@@ -86,9 +93,11 @@ export default class CommentForm extends React.Component {
               content={text}
               onBlur={() => updateStore(this.refs.editor.getContent())}
               onChange={setText}
-              onKeyDown={quickSubmit}/>
-            <input type='submit' value='Comment' ref='button'/>
+              onKeyDown={keyDown}/>
+            <input type='submit' value='Post' ref='button'
+              className={cx({enabled})}/>
             {close && <button onClick={close}>Cancel</button>}
+            <span className='meta help-text'>or press 'CMD + Enter' or 'CTRL + Enter' to post.</span>
```
levity: on Windows it's actually the Windows key, not Ctrl: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/metaKey

Maybe we should enable it for Ctrl as well.

src/components/CommentForm.js
```diff
@@ -71,10 +73,15 @@ export default class CommentForm extends React.Component {
...
-    const quickSubmit = onCmdEnter(e => {
-      e.preventDefault()
-      this.submit()
-    })
+    const keyDown = e => {
+      this.setState({enabled: this.refs.editor.getContent().length > 0})
+      onCmdEnter(e => {
+        e.preventDefault()
+        this.submit()
+      })(e)
```
levity: because onCmdEnter is curried, you can just pass the event as the second parameter instead of `()()`

src/components/CommentForm.js
robbiecarlton: Yeah. I'm guessing that was Damians intention.

src/util/textInput.js
```diff
@@ -33,5 +33,8 @@ const onKeyCode = curry((modifier, keyCode, callback, event) =>
 // use like: <input type='text' onKeyDown={onEnter(callback)}/>
 export const onEnter = onKeyCode(null, keyMap.ENTER)
 export const onCmdEnter = onKeyCode('metaKey', keyMap.ENTER)
+export const onCtrlEnter = onKeyCode('ctrlKey', keyMap.ENTER)
+export const onCmdOrCtrlEnter = (callback, event) =>
```
robbiecarlton, line 37: Let me know if you can see a point free way to write this that you'd like to use. I feel like there should be one, but rusty with this stuff.

robbiecarlton:
Good to merge now? I didn't specify windows key, because it got a little cluttered, especially with the Ctrl Key.

src/util/textInput.js
levity: Ramda appears to have "[either](http://ramdajs.com/0.22.1/docs/#either)" but i don't think there's anything in lodash/fp for this.

levity:
:+1:

PR #107, opened 2016-08-22
[In Progress] comment lists update when there's new comments

Connoropolous:


src/containers/ExpandedPostModal.js
```diff
@@ -10,19 +10,28 @@ import Post from '../components/Post'
...
+    const { commentId, post: { id, numComments }, dispatch, socket } = this.props
     dispatch(fetchComments(id, {offset: 3}))
     .then(() => {
       if (commentId) {
         const wrapper = document.getElementById(modalWrapperCSSId)
         scrollToComment(commentId, wrapper)
       }
     })
+    socket.on(`p/${id}/comment_added`, function (){
+      dispatch(fetchComments(id, {offset: numComments, refresh: true}))
```
Connoropolous: @levity could you comment on this approach of just using a message to notify that there's a new comment, and then getting the client side to fetch it? We could pass it down, but it feels like that would be weird, because then we'd have to append it into the comments array or something?

levity: I'd much rather just pull down the new comment and then append it to the
comments array. Shouldn't be any problem with doing that.

src/containers/App.js
```diff
@@ -76,7 +76,7 @@ export default class App extends React.Component {
   render () {
     const {
       children, community, dispatch, leftNavIsOpen, notifierMessages, isMobile,
-      showModal
+      showModal, route: { socket }
```
levity: `socket` seems like a good candidate to include in `childContextTypes`, since it's not specific to `ModalWrapper`.

src/routes.js
```diff
@@ -39,11 +39,17 @@ import StandalonePostEditor from './containers/StandalonePostEditor'
...
 import PageWithNav from './containers/PageWithNav'
 import TestBench from './containers/TestBench'
+import socketIOClient from 'socket.io-client'
+import sailsIOClient from 'sails.io.js'
 import { debug } from './util/logging'
 import { makeUrl } from './util/navigation'
 import { get, isEmpty, pick } from 'lodash'
 import config from './config'
 
+let io = sailsIOClient(socketIOClient)
```
levity: it feels cleaner to me to move this into App's `componentDidMount` and then pass it down through context

src/containers/ExpandedPostModal.js
Connoropolous: So, to be clear, are you saying to still perform an API call after receiving the socket event, but be sure to only fetch the comment that was just created? Or are you saying that it would be good to pass the comment object down with the websocket event? :)

src/routes.js
Connoropolous: Ok, great! The reason why I got this code up online as early as possible in my process was to make sure I was doing it in a most suitable / clean way :) thanks Lawrence.

src/containers/App.js
Connoropolous: Noted! I don't understand childContextTypes yet, but I will research it and make this work that way :)

src/containers/ExpandedPostModal.js
levity: If only subscribers to the specific comment thread get updates, we might as well send the whole comment down with the socket event, to save a round trip. If we have to broadcast to everyone, then maybe not.

Which brings up a question about security: we should make sure that only users who pass the checkAndSetPost policy can subscribe to updates for a given post. Hopefully Sails lets you use the same policies for sockets?

Connoropolous:
@levity the approach using the reducers and actions is done.

Connoropolous:
screen shot 2016-08-29 at 11 43 22 am

@levity Now the typing functionality is working, it just needs styling, but I think for that, it'd be good to have 1 of 2 things happen to finish this off...
get a primer from you on how styling in hylo works, or
have you polish that piece off and not worry about  getting me involved in that part of things. 
I am open to either, and I think my preference would be to not get involved in the css

levity:
OK, I can take it from here then.

edwardwest:
Nice work guys!

---

_Edward West_
mobile 415.225.7335
Facebook http://www.facebook.com/edwardjwest  Twitter
http://www.twitter.com/edwardwest  LinkedIn
http://www.linkedin.com/in/edwardwest

levity:
Connor, can you get this branch up-to-date, and then disable the feature, so that I can merge it into master and dark-deploy it? After that I'll work on styling.

Connoropolous:
Yep!

Connoropolous:
I've disabled the typing feature, do you want me to disable the comments coming in live as well?

src/components/CommentForm.js
```diff
@@ -87,8 +96,12 @@ export default class CommentForm extends React.Component {
...
     const placeholder = this.props.placeholder || 'Add a comment...'
+
+    const keyUp = debounce(stoppedTyping, STOPPED_TYPING_WAITTIME)
+    const startThrottled = throttle(startedTyping, MAX_TYPING_INTERVAL, { trailing: false })
     const keyDown = e => {
       this.setState({enabled: this.refs.editor.getContent().length > 0})
+      startThrottled()
       onCmdOrCtrlEnter(e => {
         e.preventDefault()
         this.submit()
```
Connoropolous, line 107: just realized that if the form is submitting it should/could call `stoppedTyping()`

levity: that makes sense to me -- feel free to commit that change

PR #112, opened 2016-09-01
Topic invite modal

robbiecarlton:
Depends on: https://github.com/Hylozoic/hylo-node/pull/168


src/reducers/tagInvitationEditor.js
```diff
@@ -1,14 +1,18 @@
 import { UPDATE_TAG_INVITATION_EDITOR, CLOSE_MODAL, SEND_COMMUNITY_TAG_INVITATION } from '../actions'
 import { partition, some } from 'lodash'
 
-export default function (state = {}, action) {
+const defaultState = {
```
robbiecarlton, line 4: curious about your thoughts on this. I don't think we do this anywhere, but it simplifies some of the code in ShareTopicModal. Any objections?

robbiecarlton:
@levity - will you take a look at this and it's counterpart in hylo-node, thanks!

src/reducers/tagInvitationEditor.js
levity: i'm ok with this

src/containers/ShareTopicModal.js
```diff
@@ -39,47 +51,187 @@ export default class ShareTopicModal extends React.Component {
...
-      dispatch(sendCommunityTagInvitation(community.id, tagName, {emails}))
+      dispatch(sendCommunityTagInvitation(community.id, tagName, {
+        users,
+        emails: emails.join(',')
+      }))
     }
 
-    return <Modal title='Invite new members' subtitle={`Invite people to join ${community.name} and follow #${tagName}`} id='share-topic' onCancel={onCancel}>
+    const copyLink = joinUrl
+      ? () => {
```
levity: can you make this a one-liner, i.e. `() => copy(joinUrl) && this.setState({copied: true})`? looks weird to have the whole thing embedded in a ternary expression otherwise

src/containers/ShareTopicModal.js
```diff
@@ -39,47 +51,187 @@ export default class ShareTopicModal extends React.Component {
...
 
     const submit = () => {
       setError(null)
 
-      let emails = parseEmailString(recipients)
-      if (isEmpty(emails)) return setError('Enter at least one email address.')
+      if (isEmpty(recipients)) return setError('Enter at least one email address or user.')
+
+      const users = flow(
+        filter(r => r.id),
```
levity: you can use lodash's string shorthand here: `filter('id')`

src/containers/ShareTopicModal.js
```diff
@@ -39,47 +51,187 @@ export default class ShareTopicModal extends React.Component {
...
+        map('id'))(recipients)
+
+      const emails = filter(r => !r.id, recipients)
```
levity: likewise this can become `omitBy('id', recipients)`

src/containers/ShareTopicModal.js
```diff
@@ -39,47 +51,187 @@ export default class ShareTopicModal extends React.Component {
...
+
     const joinUrl = get(community, 'beta_access_code')
       ? communityTagJoinUrl(community, tagName)
       : null
 
     const setError = text => dispatch(updateTagInvitationEditor('error', text))
 
-    const update = (field) => event =>
-      dispatch(updateTagInvitationEditor(field, event.target.value))
+    const update = (field) => value =>
```
levity: if you wrap this in `curry` then you can do away with the `update(...)(...)` calls later

levity:
looks good otherwise

PR #113, opened 2016-09-01
Socket.updates

levity:
Hey @Connoropolous, I'm just creating another PR with my changes so it's easy for you to review if you like. I'll make some comments inline.


src/components/Post.js
```diff
@@ -304,14 +304,14 @@ export class CommentSection extends React.Component {
     this.setState(newState)
   }
 
-  startedTyping () {
+  startedTyping = () => {
```
levity: note this syntax which creates a bound function, so we can just refer to `this.startedTyping` elsewhere instead of `this.startedTyping.bind(this)`

src/components/Post.js
```diff
@@ -259,14 +261,64 @@ export class CommentSection extends React.Component {
...
+    let newState = this.state
+    if (data.isTyping) {
+      newState.peopleTyping[data.userId] = data.userName
+    } else {
+      delete newState.peopleTyping[data.userId]
+    }
+    this.setState(newState)
+  }
+
+  startedTyping = () => {
```
levity, line 306: note this syntax which creates a bound function, so we can just refer to `this.startedTyping` elsewhere instead of `this.startedTyping.bind(this)`

src/components/CommentForm.js
```diff
@@ -11,6 +11,14 @@ import TagDescriptionEditor from './TagDescriptionEditor'
 import cx from 'classnames'
 var { array, bool, func, object, string } = React.PropTypes
 
+// The interval between repeated typing notifications to the web socket. We send
+// repeated notifications to make sure that a user gets notified even if they
+// load a comment thread after someone else has already started typing.
+const STARTED_TYPING_INTERVAL = 5000
```
levity, line 17: i increased these time limits to avoid the possibility of "flicker" if a person is taking their time typing a comment

levity:
@Connoropolous This is a pretty good first cut, so I'm going to merge it. It occurs to me though that it's possible for clients to get stuck thinking that other clients are typing, if a client starts a comment but then closes their browser or something before the "stopped typing" event is sent. If that proves to be an annoyance we can rework the functionality, e.g. by having a client keep a timestamp of the newest "started typing" event per user, and automatically considering anyone who hasn't sent a recent "started typing" event as having stopped.

Connoropolous:
Thanks for doing this! Duly noted

PR #135, opened 2016-10-12
Update local tags when new tags are created

robbiecarlton:


robbiecarlton:
depends on https://github.com/Hylozoic/hylo-node/pull/177/

robbiecarlton:
@levity - good?

src/reducers/tags.js
```diff
@@ -77,6 +78,16 @@ export const tagsByCommunity = (state = {}, action) => {
     case FETCH_CURRENT_USER:
     case CREATE_COMMUNITY:
       return mergeLeftNavTags(state, get('left_nav_tags', payload))
+    case UPDATE_POST_EDITOR:
+      const { tagDescriptions } = payload
+      let { slug } = meta
```
levity: just put `let slug` up before the switch statement rather than having the `let` here affect the assignments further on in the file

src/actions/tags.js
```diff
@@ -51,7 +51,7 @@ export function createTagInCommunity (params, slug) {
   return {
     type: CREATE_TAG_IN_COMMUNITY,
     payload: {api: true, params, path: `/noo/community/${slug}/tag/`, method: 'POST'},
-    meta: {slug, optimistic: true}
+    meta: {slug, tag: params, optimistic: true}
```
levity: If `params` is just a tag object please rename it as such -- it's pretty confusing to figure out what `params` is supposed to be here; maybe you can destructure the argument so people looking at this code know what keys to expect in it?

src/reducers/tags.js
```diff
@@ -77,6 +78,16 @@ export const tagsByCommunity = (state = {}, action) => {
...
+      if (isEmpty(tagDescriptions)) return state
+      let tags = Object.keys(tagDescriptions).map(key => ({...tagDescriptions[key], name: key, followed: true}))
+      return {...state, [slug]: mergeList(state[slug], tags, 'name')}
```
levity: it doesn't quite make sense to me that this is happening in `UPDATE_POST_EDITOR` -- shouldn't it wait until a save occurs?

levity: also, what happens in the case of a post in multiple communities?

robbiecarlton: waiting until save makes way more sense.

Don't need to update locally for communities other than the current one, because if the user goes to a different community there will be a fetchLeftNavTags that updates the store.

src/actions/tags.js
robbiecarlton: you mean destructure it and then restructure it for the api call?
like

```
export function createTagInCommunity ({ name, description, is_default }, slug) {
  return {
    type: CREATE_TAG_IN_COMMUNITY,
    payload: {api: true, params: {name, description, is_default}, path: `/noo/community/${slug}/tag/`, method: 'POST'},
    meta: {slug, tag:  {name, description, is_default}, optimistic: true}
  }
}
```

robbiecarlton: calling it params seems pretty standard, see eg updatePost and createPost. Am I missing a difference between those and this? Or are they out of date now and should also be clarified?

levity: i named them `params` because they just get sent to the API endpoint as query params. this one caught my eye because it also gets saved to `meta` with a different name. it feels like a coincidence that the tag attributes are the exact same as the query params; maybe it's as simple as naming the argument `tag` instead. or perhaps `params` in both places. ¯_(ツ)_/¯

levity: different issue: does this only work if the PostEditor was started from a community page? (See my thread in Hylo on Hylo.)

robbiecarlton: createTagInCommunity is only called when creating a tag in the tag moderation page.
the Create Tag button in post editor just updates the tagDescriptions locally until the post is saved

robbiecarlton: ie: it isn't called in post editor

robbiecarlton:
how's this?

src/actions/tags.js
levity: ahhh ok cool

levity:
:shipit:

PR #136, opened 2016-10-13
Global new messages and new threads

Connoropolous:


Connoropolous:
Consolidated APPEND_COMMENT into the ThreadsDropdown component because we can't be calling appendComment in multiple places. 

normal posts are still subscribing to a "posts/${postId}", but the receiving of "messages" (comments, but on threads) are now all coming through "users/${userId}" , whether you have that thread open or not

src/containers/ThreadsDropdown.js
```diff
@@ -25,30 +28,75 @@ const getThreads = state =>
...
+  newThread (thread) {
+    const { dispatch, currentUser } = this.context
+    dispatch(appendThread(thread))
+    dispatch(incrementUnreadThreads()) 
+    this.forceUpdate()
+  }
+
+  messageAdded (data) {
+    this.context.dispatch(appendComment(data.postId, data.message))
+    // if the threads dropdown isn't open, increment the unread count
```
Connoropolous: OR if you are on the page where you will immediately see that message. In that case, also don't increment the unread count.

Connoropolous:
the thing left to do is tracking which thread page is open, if one is, and using that to determine whether to inc. the unread count or not

Connoropolous:
also @levity it seems like there might be a conflict between what you described "new_message_count" to me as, and how it's working now... 

Weren't you describing it as something that went (permanently) to 0 if you opened the Threads popup? Right now I think if you open it, and then refresh, it will still show the same number as before because it is basing it off when threads were last read in the backend, not how long ago you opened the Threads popup. Am I making sense?

Connoropolous:
Sorry @levity , nevermind my last comment, I see now that you've marked the asana task complete and I've seen your latest commits that you've fixed what I was drawing attention to

src/reducers/people.js
```diff
@@ -130,6 +131,11 @@ const currentUserReducer = (state = null, action) => {
     case FETCH_LIVE_STATUS:
       const { new_notification_count } = payload
       return {...state, new_notification_count}
+    case INCREMENT_UNREAD_THREADS:
```
levity: can you merge my recent changes into this branch? i moved currentUserReducer into a separate file

src/containers/ThreadsDropdown.js
```diff
@@ -25,30 +28,75 @@ const getThreads = state =>
...
+  static contextTypes = {
+    dispatch: func,
+    currentUser: object
+  }
+
+  componentDidMount () {
+    const { dispatch } = this.context
+    this.socket = getSocket()
+    this.socket.post(socketUrl('/noo/threads/subscribe'))
+    this.socket.on('newThread', this.newThread.bind(this))
```
levity: you can omit the `.bind(this)` if you define the method thus:

``` js
// instead of:
newThread (thread) { ... }
// do this:
newThread = thread => { ... }
```

support for this syntax comes from the "react" Babel preset.

src/containers/ThreadsDropdown.js
```diff
@@ -25,30 +28,75 @@ const getThreads = state =>
...
+      this.socket.off('newThread')
+      this.socket.off('messageAdded')
+    }
+  }
+
+  newThread (thread) {
+    const { dispatch, currentUser } = this.context
+    dispatch(appendThread(thread))
+    dispatch(incrementUnreadThreads()) 
+    this.forceUpdate()
```
levity: this shouldn't be necessary -- if you're finding that it is, it probably means one of the reducers is modifying the existing state object when it should be replacing it.

levity:
re: consolidating logic into ThreadsDropdown: I expect that at some point, for the sake of clarity, we'll want to further tease apart this component, into one component that just responds to socket messages and one that just renders the dropdown.

levity:
anyway, looking good so far

Connoropolous:
@levity there will still be a few little tweaks after this to clean up a few edge cases, but this could possibly be merged as is

src/util/threads.js
```diff
@@ -0,0 +1,10 @@
+import { filter } from 'lodash'
+
+export const unseenThreadCount = (threads, last_viewed_messages_at) => 
+                                   filter(threads, t => threadIsUnnoticed(t, last_viewed_messages_at) && threadIsUnread(t)).length
```
levity, line 4: please just indent these 2 spaces from the left margin

src/reducers/posts.js
```diff
@@ -128,10 +128,10 @@ export default function (state = {}, action) {
       }
     case COMPLETE_POST_PENDING:
       return updatePostProps(state, id, {
-        fulfilled_at: post.fulfilled_at ? null : new Date()
+        fulfilled_at: post.fulfilled_at ? null : new Date().toISOString()
       })
     case UPDATE_POST_READ_TIME:
-      return updatePostProps(state, id, {last_read_at: new Date()})
+      return updatePostProps(state, id, {last_read_at: new Date().toISOString()})
```
levity, line 141: this date conversion would be a good thing to verify in a test

PR #138, opened 2016-10-13
Onboarding tweaks

robbiecarlton:
depends on: https://github.com/Hylozoic/hylo-node/pull/179

### DIFF
--- a/src/containers/CreateCommunity.js
+++ b/src/containers/CreateCommunity.js
@@ -38,6 +38,23 @@ import { checklistUrl } from '../routes'
 
 const merkabaUrl = 'https://www.hylo.com/img/hylo-merkaba-300x300.png'
 
+const incrementSuffix = str => {
+  const regexp = /\d+$/
+  if (!str.match(regexp)) {
+    return str + '-1'
+  } else {
+    return str.replace(/\d+$/, d => +d + 1)
+  }
+}
+
+const generateSlug = (name, oldSlug) => {
+  if (oldSlug) {
+    return incrementSuffix(oldSlug)
+  } else {
+    return name && name.toLowerCase().replace(/\s/g, '-') || ''
+  }
+}
+
 export const Topper = ({ community, showJoin }) => {
   const { name, avatar_url } = community || {}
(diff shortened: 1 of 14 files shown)


robbiecarlton:
@levity - good to merge?

One thing that I gave up on as it was hard and didn't seem important: when autopopulating the slug, if there's a match, it increments by 2 instead of 1.
So if it's trying to create hylo-1, but hylo-1 already exists, it goes to hylo-3, not hylo-2.
It's a weird artifact of timing, but I figure it doesn't actually matter. Let me know if you want me to make it right.

src/containers/CreateCommunity.js
```diff
@@ -82,7 +99,20 @@ export class CreateCommunity extends React.Component {
...
+  }
+
+  componentWillReceiveProps (nextProps) {
+    const { editedSlug, generatedSlug } = this.state
+    if (!editedSlug && get('errors.slugUsed', nextProps)) {
+      const newGeneratedSlug = generateSlug(null, generatedSlug)
+      this.setState({
+        generatedSlug: newGeneratedSlug
+      })
+      this.checkUnique('slug', newGeneratedSlug)
```
levity: I think you're incrementing by 2 instead of 1 because this is running twice. Try resetting validation before running checkUnique again.

src/containers/CreateCommunity.js
```diff
@@ -82,7 +99,20 @@ export class CreateCommunity extends React.Component {
...
-      expanded: false
+      expanded: false,
+      generatedSlug: '',
+      editedSlug: false
+    }
+  }
+
+  componentWillReceiveProps (nextProps) {
+    const { editedSlug, generatedSlug } = this.state
+    if (!editedSlug && get('errors.slugUsed', nextProps)) {
```
levity, line 106: I think you're incrementing by 2 instead of 1 because this is running twice. Try resetting validation before running checkUnique again.

robbiecarlton:
@levity - I can merge?

levity:
You can has merge

PR #140, opened 2016-10-14
Request to join fixes

robbiecarlton:


src/containers/community/CommunitySettings.js
```diff
@@ -181,6 +187,13 @@ export default class CommunitySettings extends React.Component {
       }
     }
     this.setState({expand: {...expand, [section]: open || !expand[section]}})
+    if (goToJoinRequests) {
+      setTimeout(() => {
```
robbiecarlton, line 191: curious if you can think of a better way of doing this. We have to wait for the section to expand before we can scrollTo the joinRequests, but I don't know how to do that using lifecycle methods, so I just did setTimeout.

robbiecarlton:
@levity - good to merge?
(check my comment on CommunitySettings)

src/containers/community/CommunitySettings.js
levity: I think you can check for the change in componentDidUpdate. setTimeout is fine too if it works though

levity:
:shipit:

PR #143, opened 2016-10-18
Browse topics limit

robbiecarlton:
screen shot 2016-10-18 at 2 32 39 pm

### DIFF
--- a/css/index.less
+++ b/css/index.less
@@ -946,6 +946,11 @@ ul.tags {
 
     .meta {
       margin-left: 13px;
+      a {
+        .meta;
+        margin-left: 5px;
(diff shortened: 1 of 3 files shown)


robbiecarlton:
depends on https://github.com/Hylozoic/hylo-node/pull/181

robbiecarlton:
@levity - good to merge?

levity:
I think we can just have the "Show more" link inline with the rest of the meta text. No need to distract from the main calls to action.

levity:
(Unless Damian specifically requested it that way, in which case just make the button text and border color consistent.) Otherwise, :cool:

PR #147, opened 2016-10-25
skip topics if they're all default

robbiecarlton:


src/containers/onboarding/BioPrompt.js
```diff
@@ -1,33 +1,52 @@
...
-const BioPrompt = ({ location }, { currentUser, dispatch }) => {
-  const community = find(c => c.slug === location.query.community,
-    map('community', currentUser.memberships))
+const subject = 'community'
+
+const BioPrompt = ({ location, community, tags }, { currentUser, dispatch }) => {
   const update = debounce(bio =>
     dispatch(updateUserSettings({bio})), 500)
 
+  const isNotDefault = tag =>
```
levity: i would move this logic into the body of `connect` and just pass down the boolean into the component

levity: otherwise this looks good

PR #150, opened 2016-10-27
Expanded Post opens Post Editor Modal

robbiecarlton:


levity:
this looks good; I assume the expanded post modal and the post in the list behind it both update as expected once editing is done?

robbiecarlton:
yes, it's one of those redux magic moments.

PR #154, opened 2016-11-02
copy text from create topic modal for topic moderation

Connoropolous:
relates to this issue that I logged in asana https://app.asana.com/0/140657136329201/206403127736304

makes it look like this.

screen shot 2016-11-02 at 2 55 03 pm


Connoropolous:
Was so small thought I'd rather do it than talk about it.

PR #157, opened 2016-11-04
More tests

robbiecarlton:


src/models/currentUser.js
```diff
@@ -51,6 +51,7 @@ export const newestMembership = currentUser =>
   maxBy(get('memberships', currentUser), 'created_at')
 
 export const hasFeature = (currentUser, key) => {
+  if (process.env.NODE_ENV === 'test') return true
```
robbiecarlton: How do you feel about this Lawrence? Easier than setting them per test, but maybe you can see a reason not to do this...

levity: i don't think we want to assume that the test environment has all features turned on. for example we might want at some point to test that a page renders correctly with and without a given feature enabled.

robbiecarlton: Makes sense.

src/containers/ThreadsDropdown.js
```diff
@@ -28,15 +28,16 @@ const getThreads = state =>
       return {...denormalizedPost(post, state), comments}
     }),
     compact,
-    sortBy(t => -1 * new Date(t.updated_at))
+    sortBy(t => -1 * new Date(t.updated_at),
```
levity: ??

src/components/CommentForm.js
```diff
@@ -55,7 +55,7 @@ export default class CommentForm extends React.Component {
     const { dispatch, postId, commentId, newComment, close, pending } = this.props
     if (event) event.preventDefault()
     if (!this.state.enabled || pending) return
-    const text = this.refs.editor.getContent().replace(/<p>&nbsp;<\/p>$/m, '')
```
levity, line 58: let's revert these two changes in this file since they don't do anything

src/containers/InviteModal.js
```diff
@@ -113,7 +113,7 @@ export class InviteForm extends React.Component {
     const {
       dispatch, invitationEditor, community, onClose, standalone, allowModerators
     } = this.props
-    const { currentUser } = this.context
+    const { currentUser } = this.context
```
levity: revert this please

levity:
feel free to merge once the minor fixes above are done

PR #160, opened 2016-11-11
Pass navigate to iOS app on app load

robbiecarlton:


levity:
i don't think this will solve the DM page problem though -- `navigate`, when the path is the same as the current one, gets ignored as a duplicate in `client/updateLocation.js`. i'm going to tinker around with different alternative solutions, possibly changing the duplicate-detection logic to just re-run prefetch. the open-ended solution may be to allow each component to determine how to respond by itself, i.e. by creating a custom event that components can register their own listeners on.

robbiecarlton:
ah. That makes sense. So it will actually create the DM problem for me ;)

the custom event would be `duplicateLocation` or just `navigate`?

levity:
i was thinking of something more related to re-opening the mobile app.

PR #161, opened 2016-11-12
Message thread pagination

levity:
paginate from bottom to top, from newest to oldest, not based on a
fixed offset but based on a “beforeId” parameter, which is the id of
the oldest message we already have.

also show “Loading…” when fetching the next page is in progress


levity:
hey @Connoropolous and @robbiecarlton, can one or both of you take a look at this?

robbiecarlton:
yup, will read it in the morning

robbiecarlton:
looks good

PR #179, opened 2016-12-03
Add query string notification handler

robbiecarlton:
I put it in App as I figured we might want to use it for other things. Let me know if you want me to move it.


levity:
:+1:

PR #184, opened 2016-12-12
Scroll when you send a DM

Connoropolous:
regardless of whether you've scrolled up/back through the conversation or not


src/components/MessageSection.js
```diff
@@ -36,8 +36,11 @@ export default class MessageSection extends React.Component {
   componentDidUpdate (prevProps) {
     const messagesLength = this.props.messages.length
     const oldMessagesLength = prevProps.messages.length
+    const { currentUser } = this.context
+    const latestMessage = reverse(sortBy(this.props.messages || [], 'created_at'))[0]
```
levity: rather than reversing the entire array to get the last item, let's just do `list[list.length - 1]`

levity: or `maxBy('created_at')`, which makes it a one-step process rather than having to sort then look up

levity (changes requested):
make the suggested change to `latestMessage` and remove the `.swp` file that was checked in, & we're good

src/components/MessageSection.js
Connoropolous: K!

Connoropolous:
@levity changes made

PR #186, opened 2016-12-15
Display popular skills on members page

robbiecarlton:
Depends on https://github.com/Hylozoic/hylo-node/pull/197/files

### DIFF
--- a/src/containers/People.js
+++ b/src/containers/People.js
@@ -99,13 +99,6 @@ export default class People extends React.Component {
 
     var { popularSkills } = community
 
-    /*
-    var popularSkills = [
-      'design', 'connection-making', 'systems-thinking', 'coding', 'meditation',
-      'communication', 'software-development', 'startup-advice'
-    ]
-*/
-
     // replacing hyphens with non-breaking hyphens
     popularSkills = popularSkills.map(skill => skill.replace('-', '‑'))
 
--- a/src/containers/People.js
+++ b/src/containers/People.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { prefetch } from 'react-fetcher'
 import { connect } from 'react-redux'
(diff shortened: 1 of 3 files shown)


robbiecarlton:
screen shot 2016-12-14 at 7 27 03 pm
screen shot 2016-12-14 at 7 27 21 pm

robbiecarlton:
I think something that's missing here is highlighting the selected skill when it's being searched. I've asked Alex if she has any ideas about that.

src/containers/People.js
```diff
@@ -95,7 +95,12 @@ export default class People extends React.Component {
       total, isMobile, dispatch, cacheId
     } = this.props
     const { currentUser } = this.context
-    if (error) return <AccessErrorMessage error={error}/>
+    if (error) return <AccessErrorMessage error={error} />
+
+    var { popularSkills } = community
+
+    // replacing hyphens with non-breaking hyphens
+    popularSkills = popularSkills.map(skill => skill.replace('-', '‑'))
```
levity: nice

src/containers/People.js
```diff
@@ -95,7 +95,12 @@ export default class People extends React.Component {
       total, isMobile, dispatch, cacheId
     } = this.props
     const { currentUser } = this.context
-    if (error) return <AccessErrorMessage error={error}/>
+    if (error) return <AccessErrorMessage error={error} />
+
+    var { popularSkills } = community
```
levity: if you combine this and the next line as
```js
import { map } from 'lodash/fp'
const popularSkills = map(skill => skill.replace('-', '‑'), community.popularSkills)
```
then you won't have to create the empty array in the test. in general i think it's good to have the components be forgiving in what data they expect to receive.

levity (changes requested):
Make the one suggested change and we're good

PR #190, opened 2016-12-20
fetch popular skills with graphql

levity:


src/reducers/communities.js
```diff
@@ -79,6 +80,17 @@ export default function (state = {}, action) {
       return {
         ...state, [meta.slug]: {...community, settings: payload}
       }
+    case SEND_GRAPHQL_QUERY:
+      if (meta.subject === 'popular-skills') {
```
levity, line 84: note to self based on robbie's suggestion: try out an ADD_DATA_TO_STORE-type pattern here

robbiecarlton (approved):
👍

PR #194, opened 2016-12-21
Fetch the full community record for current community on Single Post …

lorenjohnson:
…so that navigation is correct

I think the simplest solution is best in this case. Take a look, happy to approach this differently. 

NOTE: It seems like UserProfile#show / edit also has some nav issues related to not necessarily having a "not in any community context"

### DIFF
--- a/src/containers/SinglePost.js
+++ b/src/containers/SinglePost.js
@@ -143,8 +143,10 @@ const setupPage = (store, id, query, action) => {
   return Promise.all([
     communityId !== currentCommunityId &&
       saveCurrentCommunityId(dispatch, communityId, !!currentUser),
+
     // Always load the full record for the current Community record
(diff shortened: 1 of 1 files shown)


src/containers/SinglePost.js
```diff
@@ -142,7 +143,8 @@ const setupPage = (store, id, query, action) => {
   return Promise.all([
     communityId !== currentCommunityId &&
       saveCurrentCommunityId(dispatch, communityId, !!currentUser),
-
+    // Always load the full record for the current Community record
+    dispatch(fetchCommunity(communityId), {refresh: true}),
```
levity: what does `{refresh: true}` do here?

lorenjohnson: Uhh--embarrassing--that is just completely wrong.  Fixing now.

PR #195, opened 2016-12-22
Track mobile app opening and navigating from push notification

robbiecarlton:


src/containers/App.js
```diff
@@ -70,14 +72,26 @@ export default class App extends React.Component {
...
     const query = get('query', location) || {}
 
-    const version = Number(iOSAppVersion())
-    if (version < 1.7) {
+    const iOSVersion = iOSAppVersion()
+
+    const androidVersion = androidAppVersion()
+
+    if (iOSVersion > 0) {
+      trackEvent(OPENED_MOBILE_APP, {iOSVersion})
```
levity: Let's make these keys human-readable: "App Version (iOS)" and "App Version (Android)"

levity: apart from that, :+1:

PR #206, opened 2016-12-28
In feed profile completion

lorenjohnson:
Feature flagged: IN_FEED_PROFILE_COMPLETION_MODULES

Disclaimer: There is a fair bit of little fiddly refactors in this commit, for the sake of clarity I'll be sure to try and keep that to a minimum going forward, leaving anything consequential to it's own branch.


test/client/components/CommunityPosts.test.js
```diff
@@ -60,14 +71,57 @@ describe('CommunityPosts', () => {
...
+
+    it('saves bio', () => {
+      const bioTextArea = node.find(ProfileBioModule).first().find('textarea').first()
+      bioTextArea.simulate('change', {target: {value: 'test bio'}})
+      const saveButton = node.find(ProfileBioModule).first().find('button').first()
+      saveButton.simulate('click')
+      expect(store.getState().people.current.bio).to.equal('test bio')
+      // LEJ: How do I get the currentUser in context updating correctly
+      //      in this set up? Prefetch?
+      //
```
lorenjohnson: I didn't figured-out how to keep the context.currentUser up to date with store and so wasn't able to fill-out this and the subsequent two tests fully.

css/feed-modules.less
```diff
@@ -0,0 +1,73 @@
...
+
+.feed-module.profile-skills {
+  h2 {
+    font-weight: normal;
+    font-size: 22px;
+    color: @off-black;
+    margin: 0;
+    margin-bottom: 25px;
+  }
+  p {}
```
levity: remove this

src/components/ListItemTagInput.js
```diff
@@ -1,41 +1,77 @@
...
-
-ListItemTagInput.propTypes = {
-  type: string.isRequired,
-  person: object.isRequired,
-  update: func.isRequired
+    this.setState({ list: updatedList })
+    return update(type, updatedList)
+  }
+
+  remove = (item) => {
```
levity: no need for parentheses around single arguments

src/components/ListItemTagInput.js
```diff
@@ -1,41 +1,77 @@
...
-    onRemove={remove}
-    className={className}
-    filter={filter}/>
-})
-
-ListItemTagInput.propTypes = {
-  type: string.isRequired,
-  person: object.isRequired,
-  update: func.isRequired
+    this.setState({ list: updatedList })
```
levity: no spaces inside curly braces

src/components/ListItemTagInput.js
```diff
@@ -1,41 +1,77 @@
...
+    type: PropTypes.string.isRequired,
+    person: PropTypes.object.isRequired,
+    update: PropTypes.func.isRequired,
+    dispatch: PropTypes.func.isRequired
+  }
+
+  constructor (props) {
+    super(props)
+    this.state = {
+      list: props.person[props.type] || []
```
levity, line 19: it's not clear to me that local state is required at all here -- don't your local changes to `this.state.list` exactly mirror changes you're making to the app state?

src/components/ProfileBioModule.js
```diff
@@ -0,0 +1,67 @@
...
+    dispatch: PropTypes.func
+  }
+
+  constructor (props) {
+    super(props)
+    const firstName = props.person.name.split(' ')[0]
+    this.state = {
+      firstName,
+      length: 0,
+      minLength: 1,
```
levity: `minLength` and `maxLength` can just be constants defined at the top level of this file, since they don't change

src/components/ProfileBioModule.js
```diff
@@ -0,0 +1,67 @@
...
+
+  static contextTypes = {
+    dispatch: PropTypes.func
+  }
+
+  constructor (props) {
+    super(props)
+    const firstName = props.person.name.split(' ')[0]
+    this.state = {
+      firstName,
```
levity: this can just be defined as an instance variable, no need to put it in state

src/components/ProfileSkillsModule.js
```diff
@@ -0,0 +1,65 @@
...
+    dispatch: PropTypes.func
+  }
+
+  constructor (props) {
+    super(props)
+    const firstName = props.person.name.split(' ')[0]
+    this.state = {
+      tags: [],
+      valid: false,
+      firstName
```
levity: same here

src/components/ListItemTagInput.js
levity: never mind, i get what you're doing now; you're not feeding `tags` back into this component from the parent

src/containers/onboarding/SkillsPrompt.js
```diff
@@ -1,35 +1,42 @@
...
 
   const title = `Are there any skills, passions or interests you'd like to be
   known for in your community?`
 
   const subtitle = `Pick "tags" to describe yourself and to find people and
   opportunities that match your interests.`
 
+  const update = (path, value) => dispatch(updateUserSettings({[path]: value}))
+
+  const nextUrl = () => nextOnboardingUrl(location)
```
levity: i don't think this refactoring works as desired -- the property on the A tag below is getting set to this function, not the function's value

src/containers/community/CommunityPosts.js
```diff
@@ -58,42 +65,58 @@ class CommunityPosts extends React.Component {
...
+
+// Redux connection
+
+const mapStateToProps = (state, { params }) => ({
+  community: state.communities[params.id],
+  currentUser: state.people.current
 })
+
+export default compose(
+  prefetch(({ dispatch, params: { id }, query, currentUser, store }) =>
```
levity: i know you just moved this from above, didn't write it yourself, but please remove the unused arguments

src/models/currentUser.js
```diff
@@ -4,6 +4,10 @@ import { eq, flow, get, map } from 'lodash/fp'
 import { same, truthy } from './index'
 import { featureFlags } from '../config'
 
+export const hasBio = (currentUser) => currentUser.bio && currentUser.bio.length > 0
```
levity, line 7: you can just do `!!currentUser.bio` here since `!!'' == false`

test/client/components/CommunityPosts.test.js
```diff
@@ -1,9 +1,13 @@
-require('../support')
+import '../support'
```
levity, line 1: oh cool, i didn't realize this was valid

test/client/components/CommunityPosts.test.js
levity: i would skip the 'saves bio' test since you're covering it in the individual module, and just do the next three tests by modifying `currentUser` *before* you configure the store and mount the component.

levity (changes requested):

levity:
No major objections here, just some nitpicking and commentary on your tests. Should be good to go once the tests are finished.

src/containers/onboarding/SkillsPrompt.js
lorenjohnson: ah, good catch -- definitely a bug.

src/models/currentUser.js
lorenjohnson: I do slightly prefer the more explicit form I've used on this one.

test/client/components/CommunityPosts.test.js
lorenjohnson: That was a helpful suggestion / redirection--thanks! Makes better sense now.

src/components/ListItemTagInput.js
lorenjohnson: Yup

levity (approved):

PR #210, opened 2017-01-03
batch messages visually

Connoropolous:
max amount of time where messages will be batched is currently set to 5 minutes. Edward and Connor think this is a good amount because if someone sends something more than  5 minutes later its likely to be a "new thought" and thus helpful to have it visually distinctive

looks like >>
screen shot 2017-01-03 at 1 33 01 pm

### DIFF
--- a/css/threads.less
+++ b/css/threads.less
@@ -93,12 +93,12 @@
     padding-left: 30px;
 
     &.messageHeader {
-      padding-top: 5px;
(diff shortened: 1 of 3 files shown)


levity:
I'm going to tighten up the spacing between messages here, per Alex's suggestion; looks good otherwise. I'll pull down the branch locally and then merge from there.

PR #215, opened 2017-01-05
Is typing message improvements

Connoropolous:
**before**
screen shot 2017-01-05 at 4 04 15 pm

**after**
screen shot 2017-01-05 at 4 03 26 pm

**before**

screen shot 2017-01-05 at 3 27 38 pm

**after**

screen shot 2017-01-05 at 3 56 59 pm

### DIFF
--- a/css/threads.less
+++ b/css/threads.less
@@ -41,35 +41,6 @@
     background-color: #FFF;
     border-bottom: 1px solid @post-border;
     position: relative;
-
-    .newMessagesNotify {
-      position: fixed;
-      background-color: @damian-red;
-      color: #FFF;
-      border-radius: 5px;
-      bottom: 65px;
-      left: 50%;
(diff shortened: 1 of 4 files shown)


levity:
interesting -- the white border that cuts off the message text and "New Messages" button looks kind of off to me -- want to run this by Alex?

Connoropolous:
that's when you're scrolled up... yeah I can ask Alex. We could just refrain from showing the message when you're scrolled up. I thought the border also looked cheesy, at least when you're scrolled all the way to the bottom. maybe the solution is showing the border when you've scrolled up

Connoropolous:
you can see what it looks like normally, if you're at the bottom, in the second screenshot

Connoropolous:
I tried doing as Robbie suggested and just 'squeezing it in', but I guess this is why, after everything, slack squeezed it in underneath the message input

PR #222, opened 2017-01-07
Improve request to join from logged out/unregistered flow

Connoropolous:
**How it looks now**
only when the url provides an indication that they arrived here via clicking 'request to join' community while unauthed.

screen shot 2017-01-13 at 4 22 43 pm

screen shot 2017-01-13 at 4 22 32 pm

screen shot 2017-01-13 at 4 26 03 pm

### DIFF
--- a/src/containers/Signup.js
+++ b/src/containers/Signup.js
@@ -91,9 +91,13 @@ export default class Signup extends React.Component {
     const { email, next } = query
     const willCreateJoinRequest = next && 
       (next.indexOf('join=true') > -1 || next.indexOf('join%3Dtrue') > -1)
-    let subtitle = willCreateJoinRequest ? 'Once you sign up, a request to join the community will be sent.' : ''
-    subtitle = !community && !willCreateJoinRequest ? `If you're trying to join an existing community, please use the special
-        invitation link that you received from your community manager.` : subtitle
+    let subtitle = `If you're trying to join an existing community, please use the special
+        invitation link that you received from your community manager.`
+    if (willCreateJoinRequest) {
+      subtitle = 'Once you sign up, a request to join the community will be sent.'
+    } else if (community) {
+      subtitle = ''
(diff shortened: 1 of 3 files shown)


levity:
If they got to the signup page from a request-to-join button, the join request will be sent automatically once they finish signup and get redirected. So we wouldn't want to tell them to press the button again. Probably the best thing to do here is to check the value of `next` in the query params, and if it contains `join=true`, omit the subtitle entirely.

levity:
Thanks for being proactive about creating a PR in response to user feedback though! 👍

Connoropolous:
"Probably the best thing to do here is to check the value of next in the query params, and if it contains join=true, omit the subtitle entirely." 👍  will modify this to do that. I didn't know that was the behaviour

Connoropolous:
It previously looked like this, in terms of confirmation, look to the bottom left of the image

screen shot 2017-01-13 at 3 02 51 pm

src/containers/Signup.js
```diff
@@ -88,15 +88,17 @@ export default class Signup extends React.Component {
...
-    const subtitle = community ? null
-      : `If you're trying to join an existing community, please use the special
-        invitation link that you received from your community manager.`
-    const { email } = query
-
+    const { email, next } = query
+    const willCreateJoinRequest = next && 
+      (next.indexOf('join=true') > -1 || next.indexOf('join%3Dtrue') > -1)
+    let subtitle = willCreateJoinRequest ? 'Once you sign up, a request to join the community will be sent.' : ''
+    subtitle = !community && !willCreateJoinRequest ? `If you're trying to join an existing community, please use the special
```
levity: This would read better as an if-clause

Connoropolous: true true. will fix that up

levity:
Feel free to merge after making that change.

PR #226, opened 2017-01-08
Some test and test setup improvements and refactorings

lorenjohnson:
As I get a bit more up to speed on the React, Redux and Enzyme test setup in general and our particulars I'm finding some fire to clarify a few patterns. Mostly I'll be re-factoring on a couple of earlier tests I wrote for the Profile Modules to use as my own reference point for how to test going forward.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "babel-preset-stage-0": "^6.5.0",
     "bluebird": "^3.4.7",
     "bootstrap": "^3.3.5",
+    "chai-enzyme": "^0.6.1",
     "chalk": "^1.1.1",
     "cheerio": "^0.20.0",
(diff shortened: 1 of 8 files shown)


levity (commented):
looking good overall. i like the `import * as actions` pattern to support spyify.

in the places where you're mocking out `dispatch`, you can use the mockReduxStore from `test/support/mocks` instead of the real store if you don't need it to handle changes.

i see that this also contains a bugfix for contributor handling, which should be called out in the description of the PR, or perhaps extracted into a separate one.

and i still don't like spaces inside curly braces for one-liner object literals. :)

levity:
oh, i see that you have changes from #227 in here; nevermind the comment about having a separate PR.

PR #227, opened 2017-01-08
Fix: Complete Request - Contributors available across all communities…

lorenjohnson:
Allow all members of communities to which the Post belongs be selected as contributors to the request.

Requires related backend deploy: https://github.com/Hylozoic/hylo-node/pull/207

### DIFF
--- a/src/components/Post.js
+++ b/src/components/Post.js
@@ -62,7 +62,7 @@ export class Post extends React.Component {
 
   static contextTypes = {currentUser: object}
 
-  static childContextTypes = {post: object, dispatch: func, currentUser: object}
+  static childContextTypes = {post: object}
 
   getChildContext () {
     return {post: this.props.post}
--- a/src/components/Post.js
+++ b/src/components/Post.js
@@ -4,37 +4,37 @@ import { find, filter, get } from 'lodash/fp'
 const { array, bool, func, object, string } = React.PropTypes
 import cx from 'classnames'
 import cheerio from 'cheerio'
+import decode from 'ent/decode'
 import {
   humanDate, nonbreaking, present, textLength, truncate, appendInP
 } from '../util/text'
 import { sanitize } from 'hylo-utils/text'
 import { linkifyHashtags } from '../util/linkify'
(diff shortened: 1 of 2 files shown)


src/components/Post.js
```diff
@@ -62,7 +62,7 @@ export class Post extends React.Component {
 
   static contextTypes = {currentUser: object}
 
-  static childContextTypes = {post: object}
+  static childContextTypes = {post: object, dispatch: func, currentUser: object}
```
levity: is this necessary? i would think those two are already available through childContextTypes in the top-level App component

test/client/components/Post.test.js
```diff
@@ -151,68 +153,80 @@ describe('Post', () => {
...
-    node.find('.toggle').simulate('change')
-    expect(node.find('.done')).to.be.length(1)
-    node.find('.done').simulate('click')
-    expect(node.find('.contributors .person')).to.be.length(0)
-    expect(node.find('.contributors').text()).to.contain('completed')
-    expect(node.find('.toggle')).to.be.length(1)
-    mockify(window, 'confirm', prompt => true)
-    node.find('.toggle').simulate('change')
-    expect(node.find('.request-complete-heading').text()).to.contain('if this request has been completed')
+    it('requests contributors from all community associated to the post', () => {
```
levity: s/community/communities/

src/components/Post.js
lorenjohnson: Surely not necessary at all, not sure why I ended-up adding that here. Correcting.

test/client/components/Post.test.js
lorenjohnson: Got it

lorenjohnson:
I'm not understanding why CircleCI is failing the RichTextEditor tests. Everything passes locally. I tried a "Build without Cache" on Circle but same result.  Hmmm.

levity:
Since the failure pattern is different across the two attempts it could just be timeouts related to, say, network lag in the VPS that CircleCI is using. Trying again.

PR #229, opened 2017-01-09
Add alert feedback to Profile Completion Modules

lorenjohnson:


src/components/ProfileBioModule.js
```diff
@@ -41,7 +41,10 @@ export default class ProfileBioModule extends Component {
     const { value } = this.state
     const { dispatch } = this.context
     trackEvent(ADDED_BIO, {context: 'onboarding'})
-    return dispatch(updateCurrentUser({bio: value}))
+    return Promise.all([
+      dispatch(updateCurrentUser({bio: value})),
+      dispatch(notify('Thanks for sharing!', {type: 'info'}))
```
Connoropolous: all looks good code wise, the suggestion I have here is related to wording. I'd avoid the use of the word 'thanks' which installs the mental model that they just did something for hylo, or for us, when really we want them to know what they just did will benefit them. I don't think we want to come across as data collectors, and its all in the subtleties. I would suggest something simpler for the time being, just affirming, like: "Bio updated successfully." while we still have questions about the 'personality' of hylo.

src/components/ProfileSkillsModule.js
```diff
@@ -30,7 +30,10 @@ export default class ProfileSkillsModule extends Component {
   save = () => {
     const { tags } = this.state
     const { dispatch } = this.context
-    return dispatch(updateCurrentUser({tags}))
+    return Promise.all([
+      dispatch(updateCurrentUser({tags})),
+      dispatch(notify('You\'ve got skills!', {type: 'info'}))
```
Connoropolous: Same thing here for now, I would recommend 'Skills added successfully.'

lorenjohnson: 👍

src/components/ProfileBioModule.js
lorenjohnson: 👍

Connoropolous (approved):

PR #234, opened 2017-01-11
Fix a couple of bugs to do with comments on Project Requests

robbiecarlton:
Depends on https://github.com/Hylozoic/hylo-node/pull/208

### DIFF
--- a/src/components/CommentSection.js
+++ b/src/components/CommentSection.js
@@ -28,12 +28,14 @@ export default class CommentSection extends React.Component {
 
   componentDidMount () {
     const { post, post: { id }, expanded } = this.props
+    const { currentUser } = this.context
     const { dispatch } = this.context
     if (expanded) {
       this.socket = getSocket()
       this.socket.post(socketUrl(`/noo/post/${id}/subscribe`))
       this.socket.on('commentAdded', ({ parent_post_id, comment }) => {
         if (parent_post_id !== post.parent_post_id) return
+        if (comment.user_id === currentUser.id) return
         dispatch(appendComment(id, comment))
       })
     }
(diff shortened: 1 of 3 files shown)


src/reducers/posts.js
```diff
@@ -109,7 +109,6 @@ export default function (state = {}, action) {
     case CREATE_COMMENT:
       return updatePostProps(state, id, {
         follower_ids: uniq((post.follower_ids || []).concat(payload.user_id)),
-        numComments: (post.numComments || 0) + 1,
```
levity, line 112: why did this get dropped?

src/components/CommentSection.js
```diff
@@ -26,12 +27,15 @@ export default class CommentSection extends React.Component {
   }
 
   componentDidMount () {
-    const { post: { id }, expanded } = this.props
+    const { post, post: { id }, expanded } = this.props
```
levity, line 30: oh *that's* how you get both the variable itself and its contents...

robbiecarlton:
Because APPEND_COMMENT already increments the comment count, so with this
we were getting a double comment count.
I might be missing something, but it seemed like APPEND_COMMENT happens
every time CREATE_COMMENT does

levity:
Hmm, no, it shouldn't. That appears to be another bug specific to projects. It's because we never regression tested projects when we rolled out real-time comment updates and DM's...

levity:
Ah, I see. This is a bug specific to expanded posts, which include projects. APPEND_COMMENT can't be relied upon to increment comment count and so forth because the event listener isn't set up when the post isn't expanded.

levity:
One fix for this would be to also send the commenter's id down with the event and then ignore any commentAdded events from yourself.

robbiecarlton:
@levity - good now?

levity (approved):

PR #237, opened 2017-01-13
Project tagging breakup

Connoropolous:


Connoropolous:
this is really really ready for review now

src/components/PostEditor.js
```diff
@@ -110,9 +111,8 @@ export class PostEditor extends React.PureComponent {
 
   componentDidMount () {
     // initialize the communities list when opening the editor in a community
-    const { community, postEdit: { communities }, tag } = this.props
-    if (community && isEmpty(communities)) this.addCommunity(community)
-    if (tag) this.updateStore({tag})
```
levity: This can't get dropped -- it's how we pre-set the tag when you open the post editor from a tag page.

src/components/PostEditor.js
```diff
@@ -146,25 +146,20 @@ export class PostEditor extends React.PureComponent {
...
       return Promise.resolve(false)
     }
 
-    if (isEmpty(postEdit.community_ids)) {
+    if (!parentPost && isEmpty(postEdit.community_ids)) {
       window.alert('Please pick at least one community.')
       return Promise.resolve(false)
     }
 
-    if (subeditor) {
```
levity, line 163: I can see why you ended up dropping this whole branch of code, since the only thing we were validating across both events and posts was the tag... I just hope we don't end up needing it again for other fields.

src/components/PostEditor.js
```diff
@@ -285,14 +280,14 @@ export class PostEditor extends React.PureComponent {
...
-      post, postEdit, dispatch, imagePending, saving, id, defaultTags, placeholder
+      parentPost, post, postEdit, dispatch, imagePending, saving, id, defaultTags, placeholder
     } = this.props
     const { currentUser } = this.context
     const { description, community_ids, tag, linkPreview } = postEdit
     const selectableTags = uniq(compact([this.props.tag, tag].concat(defaultTags)))
     const { name, showDetails } = this.state
     const editorType = this.editorType()
-    const shouldSelectTag = editorType !== 'project'
+    const shouldSelectTag = !includes(['project', 'event'], editorType)
```
levity, line 292: I think you want to go in the other direction here and remove `shouldSelectTag` entirely, because now all types of posts should show the tag selector dropdown.

src/components/ProjectPost.js
```diff
@@ -228,18 +224,21 @@ const UndecoratedProjectPostCard = ({ post, community, comments, dispatch, isMob
...
     <Supporters post={post} simple />
     <div className='comments-section-spacer' />
-    {canComment(currentUser, post) && <CommentSection post={post}
-      comments={comments}
-      onExpand={() => dispatch(navigate(url))} />}
+    <CommentSection
+      onExpand={() => dispatch(navigate(url))}
+      isProjectRequest
+      {...{post, canComment, comments}}
+    />
```
levity: little style quibble: please put this on the same line as the last properties

src/components/ProjectPost.js
```diff
@@ -228,18 +224,21 @@ const UndecoratedProjectPostCard = ({ post, community, comments, dispatch, isMob
...
   </div>
 }
-UndecoratedProjectPostCard.contextTypes = {currentUser: object}
+UndecoratedProjectPostCard.contextTypes = { currentUser: object }
```
levity: no need for spaces inside curly braces here

src/components/ProjectPost.js
```diff
@@ -192,16 +193,15 @@ class ProjectRequest extends React.Component {
 
 const spacer = <span>&nbsp; •&nbsp; </span>
 
-const UndecoratedProjectPostCard = ({ post, community, comments, dispatch, isMobile }, { currentUser }) => {
-  const { name, user, tag, ends_at, id } = post
+const UndecoratedProjectPostCard = ({ post, parentPost, community, comments, dispatch, isMobile }, { currentUser }) => {
```
levity: parentPost seems to be unused here

src/containers/SinglePost.js
```diff
@@ -24,24 +24,32 @@ import AccessErrorMessage from '../components/AccessErrorMessage'
...
-const showTaggedPosts = post =>
-  post.tag && post.type === 'project'
+const subject = 'post'
 
 @prefetch(({ store, dispatch, params: { id }, query }) =>
   dispatch(fetchPost(id))
+  .then(action => {
+    if (action.payload.parent_post_id) {
+      return dispatch(fetchPost(action.payload.parent_post_id))
+        .then(() => Promise.resolve(action))
```
levity: Promise.resolve is redundant here; `.then(() => action)` suffices. Actually it's redundant two lines below as well since you're already inside a `then`.

Also, check for `action.error` and bail out early before assuming `action.payload.parent_post_id` exists.

src/containers/SinglePost.js
```diff
@@ -88,38 +93,43 @@ export default class SinglePost extends React.Component {
...
-          <p className='meta other-posts-label'>
-            Other posts for&nbsp;
-            <span className='hashtag'>#{post.tag}</span>
-          </p>
-          {community && <ConnectedPostList subject={subject} id={community.id}
-            omit={post.id}
-            query={{...query, tag: post.tag}} />}
+        {editing
+          ? <PostEditor post={post} parentPost={parentPost} expanded />
+          : <ShowPost post={post} parentPost={parentPost} comments={comments} />
```
levity: I think this would read better as `showPost(post, parentPost, comments)`, since ShowPost here is different than ShowPost in PostList, and the component namespace is basically global.

src/containers/ThreadPage.js
```diff
@@ -25,29 +25,21 @@ const { array, bool, func, object } = React.PropTypes
...
-  }
-
   render () {
-    const { post, error } = this.props
+    const { post, comments, error } = this.props
     if (error) return <AccessErrorMessage error={error} />
     if (!post || !post.user) return <div className='loading'>Loading...</div>
 
-    return <Thread post={post} />
+    return <Thread post={post} comments={comments} />
```
levity: not sure what this change is about -- I don't see corresponding changes in `Thread`.

src/models/currentUser.js
```diff
@@ -37,17 +37,22 @@ export const isTester = user =>
...
+  }
 }
 
-export const canComment = (currentUser, post) =>
-  currentUser && includes(post.follower_ids, currentUser.id) ||
-    !isEmpty(intersection(
-      map('community_id', get('memberships', currentUser)),
-      post.community_ids))
+export const canCommentOnPost = (currentUser, permissionsPost) => {
+  if (currentUser && permissionsPost) {
```
levity: could have just tacked `permissionsPost &&` onto the front of the chain here and put parentheses around the || part

levity: likewise with canEditPost above

src/models/post.js
```diff
@@ -50,3 +50,7 @@ export const denormalizedPost = (post, state) => ({
 })
 
 export const isMessageThread = post => get('type', post) === 'thread'
+
+export const isChildPost = post => get('parent_post_id', post)
```
levity: use !! here and below so that the values returned are true and false, not int and null

levity (changes requested):
whew! that was an epic diff.

src/components/ProjectPost.js
lorenjohnson: isProjectRequest is a bool prop and there is no corresponding value set. I could put it in the expanded props list below but I'd have to do isProjectRequest: true and it feels less better at least to me in this case :). . . Changing formatting slightly either way. I think you'll see what I mean.

levity: Oh i just meant putting the `/>` on the same line, not `isProjectRequest`

src/components/PostEditor.js
lorenjohnson: But projects and events shouldn't have the tag editor?

levity: hmm. i mean, i don't see any logical reason why they shouldn't, now that they no longer have special auto-populating tag behavior, but i suppose we can leave it off and see if anyone misses it. maybe one fewer field to look at will lead to higher completion rates.

levity (approved):

PR #246, opened 2017-01-27
give better closure to the invitation process

Connoropolous:
screen shot 2017-01-27 at 5 30 10 pm


levity:
Hmm -- this goes in the direction of making the modal single-use. If that's the case, which I'm ok with, we could just close it entirely after success and show the success message with Notifier.

Connoropolous:
what if I re-enabled it to cancel&invite if you edited the list again?

Connoropolous:
and maybe clear the list too? so they're less likely to fire off duplicates?

levity:
sure, that could work too.

Connoropolous:
I'm just going to drop this, it will be made irrelevant before long

PR #249, opened 2017-01-30
Actions refactor

lorenjohnson:
With the goal of reducing the clutter/boilerplate and clarifying our domain a bit I've created this refactoring of actions:

* Moved all actions which were in actions/index into their own category files (first pass, but it's decent) while at the same time leveraging the default actions/index.js file to include pass through action import/exports from all category action files. This will allow the simple and predictable importing of all action always from '/actions'. 

* These individual action category files can still be imported directly without utilizing the actions/index but I think until we decide to make a more decisive step towards modularizing we benefit from the simplicity and predictability of using this default index. In looking at how we had been doing things I saw a gesture towards modularization without really feeling going far enough to realize it's benefits. 
If in the future we go more modular and/or want to move these actions closer to the related reducer/s ("duck style") we will still a have clear domain segmentation in the individual action category files to start with.  Also as we build we have some healthy thinking and decisions to make in deciding how to categorize new actions. *An action creator should never be in actions/index.js going forward if we adopt this refactor.*

* In a similar step I've moved all of the action type constants from actions/index.js into /constants.js. This normalizes where we set and get constants (and I'd like to move mostly ALL text constants in here) providing clarity and some clutter-reducing options like: 

```javascript
import * as constants from './constants' 
constants.MY_CONSTANT
```
(I have not leveraged this option in this refactor as it's a style which some may not like, but I do think a central location for constants is a win over maintaining them in the individual files or functionality areas.)

I've put 8 hrs into this refactor this weekend knowing the whole time that it may not fly. I have other features pending to be working on too, this was "free time", but it's where my fire is burning right now. I really want to get all very aligned and invested together in what patterns we're using and what allows us to move more efficiently towards really clear code.  Perhaps if you're not excited about it or want a variant we can work with what is here towards that or we can just take this as a discussion piece. 

To me it's a lot of LOC change with an incremental benefit. To me this is one of a somewhat repetitive but low-risk refactors which can be done in our codebase which can improve the clarity of our domain / story told in the code and reduce any very low value noise or decisions needing to be made (e.g. which action category file to import)

Sorry if any of this is overwrought, my head has been in this for a little too long :) Let me know your thoughts everyone...

https://trello.com/c/Q0TuzjGo


levity:
I think this is a great step overall -- love the use of `export *` from index.js so that we can keep things in separate files but still keep imports simple.

Two quibbles with this approach:

- It feels better to me to keep constants which are action names separate from other types of constants.
- Thinking more long-term, if and when we have a front-end server that can break up JS into just-in-time pieces rather than a monolithic bundle (i.e. when we switch from browserify to webpack), we'll want to avoid files that import large amounts of code all at once.

But these are not significant enough at this time for me to oppose merging this.

levity (approved):

lorenjohnson:
@levity to your point about the location of the action constants, I think I agree. I'd like to move the current ./src/constants to ./src/actions/constants and we continue to handle other constants in other constants in the way we are currently. I would like to move them up there and then import them also into ./src/actions/index with the rest of the action creator categories. This will only take another 10-15 mins to accomplish. Sound good?

levity:
Works for me!

PR #252, opened 2017-01-31
Fix: Post button to enable when comments are pasted

lorenjohnson:


src/components/CommentForm.js
```diff
@@ -102,58 +102,67 @@ export default class CommentForm extends React.PureComponent {
...
+    const { currentUser, isMobile } = this.context
     const { enabled, modifierKey } = this.state
+    const editing = text !== undefined
+    const edit = () => this.setText('')
+    const placeholder = this.props.placeholder || 'Add a comment...'
 
     return <form onSubmit={this.submit} className='comment-form'>
       <Avatar person={currentUser} />
       {editing
         ? <div className='content'>
```
levity, line 149: i guess we're succumbing to the linter on this one ¯\\\_(ツ)_/¯

src/components/CommentForm.js
```diff
@@ -102,58 +102,67 @@ export default class CommentForm extends React.PureComponent {
...
+    const { postId, newComment } = this.props
+    if (!newComment) return
+    if (this.socket) this.socket.post(socketUrl(`/noo/post/${postId}/typing`), { isTyping: false })
+  }
+
+  stopTyping = debounce(this.stoppedTyping, STOPPED_TYPING_WAIT_TIME)
+
+  delaySetText = (text) => {
+    this.setEnabled(text)
+    return debounce(text => this.setText(text), 50)
```
levity: `debounce` won't work correctly like this. 

1. you have to assign it to something long-lived and reuse that reference
2. you're not calling the debounced function

you could do this:

```js
if (!this.debouncedSetText) {
  this.debouncedSetText = text => this.setText(text), 50)
}
this.debouncedSetText(text)
```

src/components/CommentForm.js
lorenjohnson: Uh yeah-- I am slowly capitulating and this is a weird case (though I don't disagree with it here). I have my sights on getting us on a CircleCI setup that runs a linter pass eventually and I generally am bringing files I touch for a task (vs a refactor) down to 0 Errors...

src/components/CommentForm.js
lorenjohnson: Duh. Yeah. I know. I kinda had a thing in the back of mind itching at me as I did it. We just went through this on another component last month. 
I'll back it up, thanks for catching this thing (again)  :)

lorenjohnson: So I confirmed that a function assigned as a class property does persist across renders. How I had it was still not right though and made a lot more sense to put the setEnabled within the debounce function either way.

levity: Oh, it definitely does. That's why `foo = debounce(() => { ... })` works. But that's not the same as `foo = () => { return debounce(...) }`

lorenjohnson: Yup, I get it. Clear enough. Thank you!

src/components/CommentForm.js
```diff
@@ -124,17 +124,17 @@ export default class CommentForm extends React.PureComponent {
...
   setEnabled = (text) => {
     this.setState({enabled: text.length > 0})
   }
 
+  stopTyping = debounce(this.stoppedTyping, STOPPED_TYPING_WAIT_TIME)
+
+  delaySetText = debounce(text => {
+    this.setEnabled(text)
+    this.setText(text)
+  }, 50)
```
lorenjohnson: So the above is how it is now.

PR #259, opened 2017-02-04
Component pods, etc

lorenjohnson:
@Connoropolous and @robbiecarlton, Lawrence and I have discussed the underlying stuff here and I want time with each of you individually or both of you to walk through it when you have a moment (30 mins-1 hr?). I will likely distill what is below into the beginnings of a React and Redux component or code style guide eventually, but I'd like to start by getting oriented together and making sure we're same-paged on what is here so far.

---

# Principles guiding this exploration and refactor
- Velocity, Collaboration, Agility and Pleasure
  - make the best practices easy and pleasurable, the bad things difficult or painful
  - clarify how to name new things and where to put them (and where existing things are)
  - invest time to research and same-page on best practices, where achievable make updates in code accordingly
- Reliability and Progressive Refactoring
  - clarify our best practices for testing 
  - zoom-out and identified where we may be missing key test coverage
  - identify and address testing friction or pain points
  - increase test coverage
  - focus on unit testing first
  - progressive refactoring: refactor whatever you touch that calls for it (use the pain principle) when completing features where appropriate test coverage exists, where it doesn't add to scope to update tests

## Related proposed practices demonstrated in this branch
- Only unit test components--expected outputs for given prop input scenarios:
  - test dumb components only, not the connected version
    - if connecting to a mock store it's probably wrong
    - * a handful of end-to-end tests which broadly cover the breadth of the connections in the application might be one helpful exception
  - enzyme shallow render, not mount (doesn't render any child nodes of component)
  - render component with each test
  - don't test "private" methods, as we mostly have integration tests this is largely solved by using mapDispatchToProps and refactoring tests accordingly
  - when testing the dumb (not connected) version of a component those tests are unit tests and go int tests/components/.. following the exact naming of the related file in the test including creating related subdirectories for "pod" components
  - check child component props being sent and use a selector for the component vs CSS class selector: 
    ```javascript
    // good
    node.find('MyComponent')
    // bad
    node.find('.my-component'))
    ```
- Decouple wrappers on connected components from component (e.g. redux connect, SSR prefetch and defer)
  - user the "pod" pattern (see src/components/Post/* for an example)
  - export all of mapStateToProps, mapDispatchToActions, fetchToState (prefetch function) functions
  - TBD: For non-trivial component connections (e.g. Post, ConnectedPostList, most all "containers", etc.) use the Ember derived "Pod" pattern currently demonstrated in Post, PostMenu and SinglePost in this branch
  - connect the component and wrap the actions outside of the component
    - if using dispatch from props or context, it's probably wrong
- If it's a component use a named function, not a const assigned ES6 arrow function
  - named functions show-up in debug stack traces vs const assigned functions are anonymous
  - it's at most only a few characters more typing in most all cases
  - for components which are the "export default" this also makes the extra export call obsolete as "export default function ..." is valid vs "export default const ..." which is not
- Use context only for currentUser, community and any well considered route provided value
- Tightly define and maintain PropTypes for component props, including especially what is required
    - if rendering the component without that prop causes an internal error in the component then the prop is required, nothing else is
    - these become very helpful in defining the baseline test fixture (not to mention in catching errors)
- General decent ref. for all of above: http://randycoulman.com/blog/2016/03/15/testing-redux-applications/
## Less important, open for debate, related stylistic things
- Don't use ES6 decorators on components for connect or prefetch
- Put the component function or class as the first thing in the file immediately following imports (with the one exception of component related CONSTANT_VALUE assignments)
- Always use the pod pattern for a connected component for all but the most trivial cases
- If it's not the component the file is named after and it's exported because it's used other places then create a new component file and name it accordingly (so Details becomes PostDetails)
- No more than one exported component per file
- No more than one connected component per file
- Helper methods are not exported for reuse from component files
- Obey the linter

## Proposals for the future
- Catch-up on linter issues, customizing settings as we see fit and then put the linter in our CircleCI deploy chain
- Discuss along what lines we might modularize things more. Currently we use a by function organization, consider moving to a feature or page/top level route modularization:
  - Refs: 
    - https://marmelab.com/blog/2015/12/17/react-directory-structure.html
    - https://jaysoo.ca/2016/02/28/organizing-redux-application
    - https://medium.com/front-end-hacking/structuring-react-and-redux-applications-255361d24f84#.52hntfz2q
- Bring styles close to associated component (e.g. Post.less.css), and/or look into modular CSS
- Bring tests close to the what is being tested (e.g. Post.js and Post.test.js all in the components directory)
- Use "duck" pattern to put actions in same file as associated reducer (and associate them). This coupling has minimal disadvantages and substantially clarifies and simplifies Redux use. Ref: https://github.com/erikras/ducks-modular-redux


src/components/CompleteRequest.js
```diff
@@ -94,10 +88,15 @@ export class CompleteRequest extends React.Component {
   }
 }
 
-export default connect((state, {post}) => {
+export function mapsStateToProps (state, {post}) {
```
levity: nitpick: `mapState`, not `mapsState`

levity (approved):
looks good except for one nitpick

src/components/CompleteRequest.js
lorenjohnson: oops, typo. good catch

PR #260, opened 2017-02-06
Hide popover was broked

Connoropolous:
the change to the css file was because the link in the popover, for the avatar, was only a portion of the size of the whole image, feeling very broken. 

is it right/good/useful to have the little notes section in the code that I added?

fixes: https://trello.com/c/7LvBvEqv/145-popovers-should-not-stay-around-after-a-link-is-clicked


src/actions/constants.js
```diff
@@ -67,13 +67,13 @@ export const INCREMENT_UNSEEN_THREADS = 'INCREMENT_UNSEEN_THREADS'
 export const JOIN_COMMUNITY_WITH_CODE = 'JOIN_COMMUNITY_WITH_CODE'
 export const LEAVE_COMMUNITY = 'LEAVE_COMMUNITY'
 export const LEAVE_COMMUNITY_PENDING = LEAVE_COMMUNITY + _PENDING
+export const LOCATION_CHANGE = '@@router/LOCATION_CHANGE' // this is the react-router action type
```
lorenjohnson: Just did poked around and this looks like the more idiomatic way to do this:
`import { LOCATION_CHANGE } from 'react-router-redux'`
So do that below when it's used and remove this pass through here.

src/reducers/index.js
```diff
@@ -55,8 +55,8 @@ import {
   FIND_OR_CREATE_THREAD,
   FINISH_LOGIN,
   HIDE_POPOVER,
+  LOCATION_CHANGE,
```
lorenjohnson: Since we don't "own" this, just do the import from react-router-redux here and don't include it in our action constants above:
`import { LOCATION_CHANGE } from 'react-router-redux'`

lorenjohnson (changes requested):
So just address that issue with the action constant and otherwise LGTM 👍  Thanks for letting me review this, it helped me warm-up to the popover change I'm looking at for mobile now,

src/reducers/index.js
Connoropolous: yes, cool, will do that

PR #261, opened 2017-02-06
linkify messages so you can follow links

Connoropolous:
@levity are there any side effects of doing this that I might not know about? I looked under the hood at some of the code, but there's actually a lot going on there. 

fixes: https://trello.com/c/ui9Q3pMu/114-2-linkify-urls-in-direct-message-threads

### DIFF
--- a/src/util/text.js
+++ b/src/util/text.js
@@ -17,13 +17,13 @@ export const truncate = (text, length) =>
 export function present (text, opts = {}) {
   if (!text) return ''
 
-  // wrap in a <p> tag
-  if (text.substring(0, 3) !== '<p>') text = `<p>${text}</p>`
+  // wrap in a <p> tag, do this be default, require opt out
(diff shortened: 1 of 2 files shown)


levity:
This wraps code in `<p>` tags, which will mess up the layout. You could change the CSS so that it still looks the same... remember to check the spacing between two consecutive messages from the same person.

Connoropolous:
@levity thanks for that. I've added an option to use present without wrapping in p tags (but it defaults to that behaviour.

src/util/text.js
```diff
@@ -17,13 +17,13 @@ export const truncate = (text, length) =>
 export function present (text, opts = {}) {
   if (!text) return ''
 
-  // wrap in a <p> tag
-  if (text.substring(0, 3) !== '<p>') text = `<p>${text}</p>`
+  // wrap in a <p> tag, do this be default, require opt out
```
levity: s/be/by

levity (approved):
looks good once you fix the typo

PR #263, opened 2017-02-07
Fix Popular skills bugs

robbiecarlton:
Fixes both of these https://trello.com/c/gxR6fRb9

### DIFF
--- a/src/containers/community/CommunityPosts.js
+++ b/src/containers/community/CommunityPosts.js
@@ -30,6 +30,7 @@ export const MIN_MEMBERS_FOR_SKILLS_MODULE = 6
 export const MIN_POSTS_FOR_POST_PROMPT_MODULE = 4
 
 const getFeedModule = (community, currentUser, moduleChoice) => {
+
   let module
 
   const showPopularSkills = community.memberCount >= MIN_MEMBERS_FOR_SKILLS_MODULE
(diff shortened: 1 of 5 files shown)


src/containers/community/CommunityPosts.js
```diff
@@ -42,10 +43,10 @@ const getFeedModule = (community, currentUser, moduleChoice) => {
 
   if (showPopularSkills && showPostPrompt) {
     module.component = moduleChoice
-    ? <PopularSkillsModule community={community} />
+    ? <PopularSkillsModule slug={community.slug} />
```
levity: i don't quite follow here -- why would this `community` not have `popularSkills`?

levity: i don't quite follow here -- why would this `community` not have `popularSkills`?

robbiecarlton: because this is getting called in componentDidMount, to avoid the server rendering warning, this community doesn't get updated when the fetchPopularSkills graphql query returns. So the popular skills are never shown.

levity: ahhhh... i see. the way we're doing this feels like a React antipattern. do you mind if i poke around this and try to refactor?

robbiecarlton: no, go ahead. Agreed it feels fishy.

levity:
ok, that's what i got -- storing the choice in state on componentDidMount, rendering nothing if the choice has not been made, and rendering the component based on the choice in render(). look good to you @robbiecarlton?

robbiecarlton:
looks great to me. I feel like we tried that at one point, and it didn't get rid of the server-side rendering warnings. But if you're not getting any warnings with this solution 👍

levity:
we have to have all four pieces. what we had before was missing "only do it in componentDidMount and render nothing when the choice hasn't yet been made"

PR #265, opened 2017-02-08
Fix long project names in project activity

robbiecarlton:
This fixes https://trello.com/c/oluOqQv7

screen shot 2017-02-08 at 2 42 10 pm

Do we want to do anything about the project name in the post meta as well? Doesn't exactly break anything, but it's a little much. I took a shot at it and it's tricky and didn't want to go further down that rabbit hole unless we feel it's worthwhile.

### DIFF
--- a/css/posts.less
+++ b/css/posts.less
@@ -1313,6 +1313,12 @@ ul.posts {
     border-bottom: 2px solid @faint-gray-border;
     height: 80px;
     line-height: 80px;
+    display:flex;
+    flex-direction: row;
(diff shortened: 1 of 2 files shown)


levity:
Hmm -- I'm inclined to just omit it in the post meta since it's redundant with the header.

levity:
I'd say if that's a straightforward change then toss it in here, otherwise never mind.

robbiecarlton:
yeah, its simple

PR #266, opened 2017-02-09
refactor BrowseTopicsModal

levity:
I took the opportunity to try out @lorenjohnson’s new pattern while
working on https://trello.com/c/UPaNWQwr


src/components/BrowseTopicsModal/connector.js
```diff
@@ -0,0 +1,32 @@
...
+  return {
+    ...connectedListProps(state, {subject, id: community.slug}, 'tags'),
+    followedTags: getFollowedTags(community, state)
+  }
+}
+
+export function mapDispatchToProps (dispatch, { onboarding }) {
+  return {
+    actions: bindActionCreators({
+      fetchTags: (community, offset = 0) =>
```
lorenjohnson, line 19: On further thought, why not just modify the fetchTags action to accept a community key (or post, etc) instead of a subject+id (or creating a new fetchTagByCommunity action). I'm seeing that we use fetchTags in only 2 other components... Either way, I think we should push the intention of simplifying our consumption of fetchTags either into that action or a new action so that the "problem" gets presented at the right layer where it should be solved now or later.

src/components/LeftNav.js
```diff
@@ -57,7 +59,7 @@ const TopicList = ({ tags, slug }, { dispatch }) => {
     {!isEmpty(followed) && followed.map(tag =>
       <TagLink name={tag.name} key={tag.name} highlight={tag.new_post_count} />)}
     {slug && <li>
-      <a onClick={() => dispatch(showAllTags(slug))} className='browse-all'>
+      <a onClick={() => dispatch(showAllTags(community))} className='browse-all'>
```
lorenjohnson: I think we should connect this component, or better, pass a showMoreTopicsClickHandler function as a prop from a connected component higher up. 

To move incrementally though we can start by connecting and mapDispatchToProp'ing in all cases where we have been using dispatch within the component. At that point we have a unit testable component with the right API in place and later we can go back and address any unnecessary state coupling in our architecture in addressing these small components being connected to the store case by case. 

So for now in this case in the mapDispatchToProps bindActionCreators call I'd do this:

`showMoreTopicsClickHandler: (community) => showAllTags(community)`

...and then obviously we'd end up with this on the link:

`onClick={this.props.showMoreTopicsClickHandler}`

test/client/components/BrowseTopicsModal.test.js
```diff
@@ -1,11 +1,10 @@
 require('../support')
```
lorenjohnson: Move this into test/components now that it's only a component unit test.

test/client/components/BrowseTopicsModal.test.js
```diff
@@ -59,42 +58,28 @@ describe('BrowseTopicsModal', () => {
...
     beforeEach(() => {
-      currentUser = {
-        memberships: [
-          {...community, community_id: '1'}
-        ]
+      props = {
+        onboarding: true,
+        community,
+        tags
       }
```
lorenjohnson: Are these 3 props the minimum required vales for a render? Make sure to indicate them on the component. If they are not required then consider factoring the test to only begin with what is required for render.

test/client/components/BrowseTopicsModal.test.js
```diff
@@ -59,42 +58,28 @@ describe('BrowseTopicsModal', () => {
...
+
     beforeEach(() => {
-      currentUser = {
-        memberships: [
-          {...community, community_id: '1'}
-        ]
+      props = {
+        onboarding: true,
+        community,
+        tags
```
lorenjohnson: Are these 3 props the minimum required vales for a render? Make sure to indicate them on the component. If they are not required then consider factoring the test to only begin with what is required for render.

test/client/components/BrowseTopicsModal.test.js
lorenjohnson: Move this into test/components now that it's only a component unit test.

src/components/ModalWrapper.js
```diff
@@ -0,0 +1,129 @@
...
+    if (this.props.bottom) window.removeEventListener('scroll', this.lockScrolling)
+  }
+
+  render () {
+    const { type, params, top } = this.props
+    const { dispatch } = this.context
+    if (!type) return null
+
+    let modal, clickToClose
+    const close = () => dispatch(closeModal())
```
lorenjohnson: Connect this component with this in mapDispatchToProps, bindActionCreators:

`closeClickHandler: () => closeModal()`

src/components/BrowseTopicsModal/component.js
```diff
@@ -0,0 +1,126 @@
...
+        </ul>}
+      {onboarding && <div className='footer'>
+        <A className='button' to={nextUrl}>Next</A>
+      </div>}
+    </Modal>
+  }
+}
+
+const TagRow = ({ tag, community, followed, onClick, onFollow }, { isMobile }) => {
+  const { id, name } = tag
```
lorenjohnson: used a named function

src/components/BrowseTopicsModal/component.js
```diff
@@ -0,0 +1,126 @@
...
+    </p>}
+    {!isEmpty(owner) && <span className='meta'>
+      created by <A to={userUrl(owner)}>{owner.name}</A> {humanDate(created_at)}
+    </span>}
+    {isMobile && controls}
+  </li>
+}
+TagRow.contextTypes = {isMobile: bool, dispatch: func}
+
+const TagRowControls = ({ followed, follower_count, name, onFollow }) => {
```
lorenjohnson: ...named function

lorenjohnson (approved):
Generally LGTM, what I have commented on is all related to moving any components we touch, unless something urgent / a hot bug fix, into being just components using mapDispatchToProps. I do think we should pretty strictly do that as our first pass, even though this will land us with more officially "connected" components than before (though using dispatch from context is essentially connecting it). Take a look.

PR #267, opened 2017-02-09
Comment image fixes

robbiecarlton:
This addresses: https://trello.com/c/rMUHlqIK


robbiecarlton:
screen shot 2017-02-10 at 3 23 43 pm

robbiecarlton:
@levity - good to merge?

robbiecarlton:
ok. dispatch moved out of components.

src/components/CommentImageButton.js
```diff
@@ -1,39 +1,53 @@
...
+  currentUser: object.isRequired
 }
 
 const mapStateToProps = (state, props) => ({
   pending: get([UPLOAD_IMAGE, 'subject'], state.pending) === 'comment-image'
 })
 
-export default connect(mapStateToProps)(CommentImageButton)
+export function mapDispatchToProps (dispatch, { post }) {
+  const actions = bindActionCreators({
```
levity: There's a shorthand format that `connect` also accepts:

```js
export const mapDispatchToProps = {createComment, notify, uploadImage}
```

it'll assume each value in the object is an action creator and bind them all.

levity: although I'm not sure it works if they're nested under `actions` though.

levity (approved):

levity (changes requested):
I think you also need to disable the Send button in messages when the image upload is pending.

src/components/CommentImageButton.js
lorenjohnson: From what I can tell the shorthand format for mapDispatchToProps will not let us wrap things under a key.

In final analysis I think in nearly all cases it's probably better to not wrap them under an "action" key. 

Maintaining a propTypes spec will help to keep the component API grok'able on a glance in cases where there are a larger number of functions coming in ...

levity (approved):

PR #270, opened 2017-02-10
Refactor round 2

lorenjohnson:
This one I admit caught me by surprise, it began as a fix for no Popovers on mobile ticket and I ended-up taking another deeper dive with the intention of exploring, setting-in a little more firmly the patterns we've adopted as of the first round. I think I can decelerate from here to let things settle, but we'll all have more to consider while we do it. 

My opinion is that everything here is better than before I touched it and generally going in the right direction, with one possible exception:

* The pod'ification of a few smaller components I feel iify about really benefiting as much from the pattern. I REALLY like thinking about just the component and not having the extra connection noise when doing so, but when it's only one method I think that the extra noise of the pod file setup could be a wash at best.  So I may collapse again Avatar and PostHeader, but for now I'd like to leave them and see how it feels for all of us for a bit. . . 

* (re)Normalize <Avatar /> and <ClickCatcher /> favoring clarity and testability
* Migrated the following components into pods: Avatar, ClickCatcher, Comment, LinkedPersonSentence, Popover, PostHeader, PostList, ProjectPost
* Split PersonCards and PersonCard (pod), ProjectPost and ProjectPostCard (both now pods)
* Split PostDetails out of Post for clarity and testability
* Updated various enzyme mounted tests to shallow rendering
* Add and clarify various tests

Again, sorry to draw outside the lines--off schedule and without clear consent-- but I think setting and exploring some more clear patterns for us is pretty key before we jump into quickly building new stuff or possibly stepping down into maintenance mode on this code base.


test/components/Avatar/component.test.js
```diff
@@ -0,0 +1,55 @@
...
+    expect(wrapper.find('div.avatar').length).to.equal(1)
+    expect(wrapper.find('Link').length).to.equal(1)
+  })
+
+  it('shows change image button when showEdit is set', () => {
+    const wrapper = renderComponent({showEdit: true})
+    expect(wrapper.find('Connect(ChangeImageButton)').length).to.equal(1)
+  })
+
+  it('shows link when isLink is false', () => {
```
levity: i think you mean "doesn't show link" here

test/components/Post/component.test.js
```diff
@@ -71,24 +67,22 @@ describe('Post', () => {
...
-  //   // LEJ: Need to make a window mock for this unit testing
-  //   //      context or move FEATURE_FLAGS to a "features" prop
-  //   window.FEATURE_FLAGS = { CONTRIBUTORS: 'on' }
-  //   const props = {post: {tag: 'request'}}
-  //   const wrapper = renderComponent(props)
-  //   expect(wrapper.find('Connect(CompleteRequest)').length).to.equal(1)
-  //   delete window.FEATURE_FLAGS
-  // })
+  it('should have the complete request feature if enabled and the post is a request', () => {
+    process.env.FEATURE_FLAG_CONTRIBUTORS = 'on' // window.FEATURE_FLAGS = { CONTRIBUTORS: 'on' }
```
levity: let's do the env changes in before and after blocks -- this way, if the test fails, the env var doesn't get deleted, and it could affect other tests

test/components/PostList.test.js
```diff
@@ -78,13 +43,55 @@ const posts = [
...
-    expect(doc('.post').length).to.equal(3)
-    expect(doc('.project-summary').length).to.equal(1)
-    expect(doc('.event-summary').length).to.equal(1)
-    expect(doc('.my-module').length).to.equal(1)
+describe('<PostList />', () => {
+  it('should shallow render', () => {
+    const wrapper = shallow(<PostList posts={posts} />)
+    expect(wrapper.find('ShowPost').length).to.equal(4)
+    expect(
+      wrapper.find('ShowPost').map((postWrapper) => postWrapper.dive().name())
```
levity, line 51: :cool:

src/components/Avatar/connector.js
```diff
@@ -0,0 +1,10 @@
+import { connect } from 'react-redux'
+import { showPopoverHandler } from '../../actions'
+
+export function mapDispatchToProps (dispatch) {
+  return {
+    onMouseOver: showPopoverHandler(dispatch)
```
levity: the naming feels a bit off here -- calling it `onMouseOver` suggests to me that it will always be the mouseover behavior, but it's controlled by the `showPopover` bool. maybe just keep calling it `showPopoverHandler` through the component?

levity (changes requested):
Looks sound overall. For changes, see my handful of comments, and also finish the stubbed/commented-out tests. Going forward, let's try to keep the diffs smaller so they're easier to review.

src/components/Avatar/connector.js
lorenjohnson: Fully agree-- good catch. I was in a tug of war with that one and in a few other cases, then settled on the same conclusion, but neglected to changing this one back.

test/components/PostList.test.js
lorenjohnson: Right? With a little "dive" now and then shallow rendering can cover just about any case.

test/components/Post/component.test.js
lorenjohnson: Oh yeah--good point.

lorenjohnson:
I've addressed what you caught and I think this is ready to merge. One note re. pending tests: I will sometimes purposely leave a test in pending state to bookmark it for later attention.

Thank you so much BTW! I'm sorry it was a beast again, I will definitely make things more manageable going forward now with these patterns getting set in place.

PR #271, opened 2017-02-10
basic optimism for comments and messages

Connoropolous:
- [x] add/fix tests
- [x] get feedback on look/feel
- [x] add handler for comment image create error


src/components/ImageWithFallback.js
```diff
@@ -0,0 +1,39 @@
+import React, { PropTypes } from 'react'
+
+export default class ImageWithFallback extends React.Component {
+ static propTypes = {
+    preferredSrc: PropTypes.string.isRequired,
+    fallbackSrc: PropTypes.string.isRequired,
+    useClass: PropTypes.string
```
levity: let's call this className to be consistent with React convention

src/components/ImageWithFallback.js
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor (props) {
+    super(props)
+    this.state = {
+      imageStatus: false
+    }
+  }
+
+  handleImageLoaded = () => {
+    this.setState({imageStatus: true});
```
levity: a semicolon snuck in here

src/components/ImageWithFallback.js
```diff
@@ -0,0 +1,39 @@
...
+ static propTypes = {
+    preferredSrc: PropTypes.string.isRequired,
+    fallbackSrc: PropTypes.string.isRequired,
+    useClass: PropTypes.string
+  }
+
+  constructor (props) {
+    super(props)
+    this.state = {
+      imageStatus: false
```
levity: maybe this should be called `preferredImageLoaded`

src/actions/comments.js
```diff
@@ -43,12 +44,23 @@ export function appendComment (postId, comment) {
...
+export function createComment ({ postId, text, tagDescriptions, imageUrl, userId }) {
   const params = {text, tagDescriptions, imageUrl}
+  const comment = {
+    id: uniqueId(`post${postId}_`),
+    user_id: userId,
+    thank_ids: []
+  }
+  if (text) {
+    comment.text = text
+  }
```
levity: put `else` on the same line as the closing brace from `if`

src/actions/comments.js
```diff
@@ -43,12 +44,23 @@ export function appendComment (postId, comment) {
   }
 }
 
-export function createComment ({ postId, text, tagDescriptions, imageUrl }) {
+export function createComment ({ postId, text, tagDescriptions, imageUrl, userId }) {
   const params = {text, tagDescriptions, imageUrl}
+  const comment = {
```
levity: let's call this `optimisticComment` or `tempComment`? it's a little confusing otherwise to see in the reducer that the comment creation has successfully completed, and so we should _remove_ something matching `comment.id`

src/components/Comment/component.js
```diff
@@ -75,34 +77,32 @@ export default class Comment extends React.Component {
...
-      <a name={`comment-${comment.id}`} />
+      {notPending && <a name={`comment-${comment.id}`} />}
       <Avatar person={person} showPopover />
       <div className='content'>
         {image && <div className='text'>
           <A to={`/u/${person.id}`} className='name'>{person.name}</A>
         </div>}
         {image && <a onClick={() => showImage(image.url, location.pathname, isMobile)}>
-          <img className='thumbnail' src={image.thumbnail_url} />
+          <ImageWithFallback useClass='thumbnail' preferredSrc={image.thumbnail_url} fallbackSrc={image.url} />
```
levity: If we use this for all comments, doesn't that cause the large image size to be loaded for all existing comments with images? we'd definitely want to avoid that...

src/components/MessageForm/component.js
```diff
@@ -32,10 +32,19 @@ export default class MessageForm extends React.Component {
...
+    const userId = currentUser.id
     const { text } = this.state
 
-    createComment({postId, text})
-    .then(({ error }) => error || trackEvent(SENT_MESSAGE))
+    createComment(text, userId)
+    .then(({ error }) => {
+      if (error) {
+        this.setState({text})
+      }
```
levity: put `else` on same line as closing brace

src/components/MessageSection.js
```diff
@@ -24,7 +24,7 @@ function createMessageList (messages) {
       messageDate = new Date(m.created_at)
       diff = Math.abs(headerDate - messageDate)
       greaterThanMax = Math.floor(diff / 60000) > MAX_MINS_TO_BATCH
-      isHeader = greaterThanMax || m.user.id !== currentHeader.user.id
+      isHeader = greaterThanMax || m.user.id !== currentHeader.user.id || !m.created_at // last case means pending
```
levity: will this cause visual flicker if the message wasn't otherwise supposed to have a header?

levity (changes requested):

src/components/ImageWithFallback.js
Connoropolous: ah, I thought there might be a conflict, ok

src/components/ImageWithFallback.js
Connoropolous: yeah, I'll rename for more clarity

src/actions/comments.js
Connoropolous: yeah, that's what I had it as initially, but then for some reason when I was talking with Robbie I decided to name it this. I think looking back now I agree with you though, so I'll do that

src/actions/comments.js
Connoropolous: sorry about the stylistic things, I've been coding on C9 still because running Atom on my computer, oddly, just doesn't go that well for me. I'm getting that sorted soon!
Also, I was going to run js-standard over these files, I'll do that next time. Just when I thought I got it right too, cause I've been trying to actually add it in to my memory too

src/components/Comment/component.js
Connoropolous: right... I have some complexity here though, because the code can't reason about what's a newly created saved comment, and a previously already saved comment. I may have to add in some kind of manual flag to distinguish I guess

src/components/MessageSection.js
Connoropolous: Yes, but otherwise we miss out on the 'sending...' feature... maybe that doesn't matter. but it will be kind of a bad UX if the comment creation fails, in this instances where its not a header

Connoropolous:
@levity all changes made here

src/components/Comment/component.js
```diff
@@ -75,34 +77,32 @@ export default class Comment extends React.Component {
...
-      <a name={`comment-${comment.id}`} />
+      {notPending && <a name={`comment-${comment.id}`} />}
       <Avatar person={person} showPopover />
       <div className='content'>
         {image && <div className='text'>
           <A to={`/u/${person.id}`} className='name'>{person.name}</A>
         </div>}
         {image && <a onClick={() => showImage(image.url, location.pathname, isMobile)}>
-          <img className='thumbnail' src={image.thumbnail_url} />
+          <ImageWithFallback className='thumbnail' preferredSrc={image.thumbnail_url} fallbackSrc={image.url} onlyUsePreferred={!fromTemp} />
```
levity: Maybe instead of adding another attribute, you can just omit the fallback when you know the thumbnail is going to be available, and then in the ImageWithFallback component, you can skip the fallback behavior if the fallback is not defined. I think that would make this component a bit easier to reason about and drop in wherever we want to have this "flicker-free" behavior in the future.

levity (changes requested):

Connoropolous:
Right, makes sense.

levity (approved):

PR #272, opened 2017-02-14
No popovers on mobile

lorenjohnson:
1) Carrying forward with my intent to bring up whatever I touch I went ahead and pod'ified App and added very basic render tests, as well as coverage for this ticket.

    NOTE: As dispatch is no longer used internally in the component I've changed dispatch from being passed from props to childContext to context to childContext.

2) Handling the blocking of mobile rendering of the popover in App where the popover is rendered rather than adding an isMobile to every call of handle / showPopover. 

    This perhaps could be handled in the popover reducer, but I didn't see off-hand how to access isMobile in the state tree at the point of the popover reducer nor was I confident this would be idiomatic or better than what I've done. Happy to change it to any other way if there is a better way or a reason to not do it the way I've chosen. 

https://trello.com/c/cO2MOx5O

### DIFF
--- a/src/containers/App/component.js
+++ b/src/containers/App/component.js
@@ -92,7 +92,7 @@ export default class App extends React.Component {
   render () {
     const {
       children, community, leftNavIsOpen, notifierMessages, openModals,
-      popover, currentUser, location, removeNotification
+      popover, currentUser, location, isMobile, removeNotification
     } = this.props
 
     const classes = cx({
@@ -104,6 +104,7 @@ export default class App extends React.Component {
 
     const { pathname } = location || {}
     const showIntercomButton = pathname && !pathname.startsWith('/t/')
+    const showPopover = !isEmpty(popover) && !isMobile
 
     return <div className={classes}>
       {children}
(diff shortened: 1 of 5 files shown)


test/client/components/ShareTopicModal.test.js
```diff
@@ -2,8 +2,8 @@ require('../support')
 import React from 'react'
 import { mount } from 'enzyme'
 import { configureStore } from '../../../src/store'
-import App from '../../../src/containers/App.js'
```
lorenjohnson, line 5: I am assuming this ".js" thing was a legacy thing and no longer necessary even in the server context? It of course breaks the App/index.js default thing relied upon for the pod.

levity: yep

src/containers/App/component.js
```diff
@@ -64,12 +46,13 @@ export default class App extends React.Component {
   }
 
   getChildContext () {
-    const { dispatch, currentUser, isMobile, location } = this.props
-    return {dispatch, currentUser, isMobile, location}
+    const { currentUser, isMobile, location } = this.props
+    const { dispatch } = this.context
```
levity: App doesn't have dispatch in its context -- it's the component that puts it into context for everyone else -- you have to get it from props.

levity (changes requested):
fix dispatch and you're good

src/containers/App/component.js
lorenjohnson: I get that completely now, but for some reason I thought that Redux put dispatch into context for anything connected. My bad. Will fix.

lorenjohnson: (It makes me very uneasy that non of our tests caught this issue... but the required PropType for dispatch will at least give us that warning once I put this right again.)

levity (approved):

PR #276, opened 2017-02-15
Close modals on location change

robbiecarlton:
This addresses https://trello.com/c/WnKFOtFj
and so much more!


src/actions/constants.js
```diff
@@ -1,3 +1,7 @@
+import { LOCATION_CHANGE } from 'react-router-redux'
+
+export { LOCATION_CHANGE }
```
levity: if i'm not mistaken, you can just do
```js
export { LOCATION_CHANGE } from 'react-router-redux'
```

Connoropolous: in the other file where we reference location change, we just import it directly in that file from react-router-redux

Connoropolous: oh I see, nvm, I hadn't seen the whole diff, sry

levity (approved):

PR #277, opened 2017-02-16
image modal wasn't closing on clicking the X

Connoropolous:


robbiecarlton:
Hey @Connoropolous, nice catch!
Weird that that wasn't plugged in. Not sure if it was a revert or I just never actually clicked the X. Clicking outside the image always worked.

I've opened another PR that has an alternate fix to this.
https://github.com/Hylozoic/hylo-redux/pull/278
This is how it was supposed to work (also fixes related issue in BrowseTopicsModal).

Not sure which I prefer. I think in a vacuum, yours makes sense. Mine is more in line with what's already happening in the (admittedly baroque) Modal code. 
Plus one fewer connected component, if we care about that...

Connoropolous:
@robbiecarlton gotcha. lets just go with yours, that's fine :)

PR #278, opened 2017-02-16
Fix image modal close button

robbiecarlton:
alternate fix to: https://github.com/Hylozoic/hylo-redux/pull/277


levity:
this is my fault -- checked in an ill-considered refactoring. the idea was to build the close behavior into Modal directly but i overlooked that many modals also need to call upon that behavior after other inputs.

PR #286, opened 2017-03-01
fix invite styles for mobile especially

Connoropolous:
We were trying to do some special stuff for the web layout as distinct from the mobile layout but there were some holes in the idea, so I think the shortest route to utility and cleanliness here is just to resort back to some simplistic positioning on these elements. 

I hid the CSV import when you're on mobile. I don't think there's a use case for it.

screenshot 2017-02-28 at 11 40 36 pm
screenshot-hylo-redux-connoropolous c9users io 2017-02-28 23-25-00

### DIFF
--- a/css/communities.less
+++ b/css/communities.less
@@ -58,6 +58,15 @@
 }
 
 #community-invite-form {
+  .community-invite-form-header {
+    font-size: 20px;
+    font-weight: 500;
+    padding-left: @modal-row-side-padding;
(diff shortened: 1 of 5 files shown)


Connoropolous:
one can see how broked it was here: https://trello.com/c/SE8Ftceu/234-invite-page-on-mobile-bad-broken-styling

levity (approved):

PR #287, opened 2017-03-01
fix the image modal closing issue

Connoropolous:
https://trello.com/c/0Hn7j1f4/210-some-parts-of-the-backdrop-behind-an-image-modal-cannot-be-clicked-to-close-it

### DIFF
--- a/src/containers/ImageModal.js
+++ b/src/containers/ImageModal.js
@@ -15,7 +15,6 @@ export default class ImageModal extends React.Component {
 
   handleClick (event) {
     const { onCancel } = this.props
-    event.persist()
     if (event.target.nodeName !== 'IMG') onCancel()
   }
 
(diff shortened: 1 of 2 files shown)


src/containers/ImageModal.js
```diff
@@ -13,6 +13,12 @@ const leftOffset = () => {
 export default class ImageModal extends React.Component {
   static propTypes = {url: string, onCancel: func}
 
+  handleClick (event) {
+    const { onCancel } = this.props
+    event.persist()
```
levity: what's the purpose of `persist` here?

Connoropolous: it had to do with react synthetic pooling, but I think that may actually have only been an issue doing console.logging
https://facebook.github.io/react/docs/events.html

Connoropolous: I will test if its actually needed

PR #292, opened 2017-03-15
Fix realtime comments on posts

Connoropolous:
most of the time this line
`if (parent_post_id !== post.parent_post_id) return`
was executing as `null !== undefined` which returns true which forces the function to return here, instead of displaying the comment. 
@robbiecarlton I tried to reason through this thinking about parent posts, but couldn't find a scenario to be concerned about. am I forgetting something?

goes with https://github.com/Hylozoic/hylo-node/pull/228

### DIFF
--- a/css/posts.less
+++ b/css/posts.less
@@ -363,6 +363,7 @@ ul.posts {
     .typing {
       color: @faint-gray-text;
       padding-top: 10px;
+      padding-left: @post-content-left-margin;
       > div {
         display: inline-block;
       }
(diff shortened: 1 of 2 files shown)


levity:
I recall that we put this in place to prevent comments on child posts from showing up in parent posts -- but as I go through the logic now, this doesn't appear to be an issue? @robbiecarlton

robbiecarlton:
Hmm. Tried to remember what this does specifically, but not sure. This is why tests are good :/ 
It did successfully fix a bug though. I remember that much. Sorry that's not more helpful.

Connoropolous:
I believe this would have been under the circumstances where post header and project post were inverted and then I flipped them back. This event subscription only ever happens now on an expanded post, and it always subscribed to the events for its own post ID so the comment being on the post itself should at this point always be guaranteed

levity (approved):

PR #293, opened 2017-03-15
fix circumstance where newMessageCount wouldn't increment right

Connoropolous:
the reason we're unable to rely on post.updated_at for the latest comment time in these circumstances is the because the Post record is updated asynchronously by the worker now, after the comment has already been fully created, and the websocket events sent

this PR also handles an improvement to the way it handles things after socket reconnection, so that you will see a newMessageCount if there is one

### DIFF
--- a/src/containers/ThreadsDropdown.js
+++ b/src/containers/ThreadsDropdown.js
@@ -113,12 +113,13 @@ export default class ThreadsDropdown extends React.Component {
       this.initialFetchThreads(true)
     } else {
       dispatch(appendComment(postId, message))
-      const updatedThreads = map(t => t.id === postId ? { ...t, updated_at: message.created_at } : t, threads)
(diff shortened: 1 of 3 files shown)


src/containers/ThreadsDropdown.js
```diff
@@ -109,16 +109,23 @@ export default class ThreadsDropdown extends React.Component {
 
     if (this.state.open || thisThreadOpen) return dispatch(appendComment(postId, message))
 
-    if (!threads.length) { // or you don't have the thread in question?
```
Connoropolous, line 112: As of now, the comment/question here should never be the case, because if you're connected to the socket it will have refreshed the thread list. If you're not connected to the socket this event can't fire

levity:
Can you write a few tests for this? Even though it's in the old code base it'll be a useful reference as we build out the new one.

Connoropolous:
yep, do you want me to do that, then merge? I've got some momentum  on a different track atm

levity:
yep!

PR #295, opened 2017-03-23
use horizontal scrolling for overflow of sections on people profile

Connoropolous:
fixes https://trello.com/c/9RmST07w

see how it looks/works here: https://youtu.be/0V7McjosifY

also fixes a small bug where the active item wasn't getting the proper class attached to it to show as active

this works similar to how the header on github repos works on mobile

### DIFF
--- a/css/people.less
+++ b/css/people.less
@@ -246,6 +246,7 @@
     .mobile({
       white-space: nowrap;
       overflow-x: scroll;
+      padding: 0 10px;
       a {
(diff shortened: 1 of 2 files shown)


levity:
Oh, neat solution! A little more padding around the far left and right
would be good.

Connoropolous:
yes, you're right

levity:
Feel free to merge after that

